PR #2, opened 2020-11-17
Allow metadata to be passed into the engine

brenetic:
In preview mode the service metadata will not be available on startup therefore we need the ability to pass into the service object. This will be the case in the editor.

In the runner the service metadata should be there at startup so we can read it in and keep it in memory. The location
'/spec/fixtures/service.json` is a temporary one until we decide how the publisher will pass the metadata into the runner's container.

The fixture lives in the fb-runner repo.

https://trello.com/c/QOHT8XOs/1076-create-barebones-fb-runner


tomas-stefano (approved):
🚀

PR #7, opened 2020-11-19
Move helper method to controller

tomas-stefano:
## Context

The editor in the test environment was not booting:

```
/usr/local/bundle/bundler/gems/fb-metadata-presenter-9b8c0380ad68/lib/metadata_presenter/engine.rb:11:in `block (2 levels) in <class:Engine>': undefined method `helper' for ActionController::API:Class (NoMethodError)
```

This PR fixes the issue. The reason is that `helper` method is not available in the initializer. Since the `helper` is available in the controller, moving to controller makes everything works as expected.


brenetic (approved):
🚀

PR #10, opened 2020-12-01
Save the data into the session

tomas-stefano:
## Context

This PR makes the possibility to save in the session possible (for the editor, runner and the dummy app).

## Disclaimer

I had to change the way we process the answers through the form.

Some problems are using the `request.referer`:

*  The editor uses '/services/:service_id/preview' as base URL.

  ```
  The problem redirecting to '/text-first' is that this url doesn't
  exist on the editor (only '/services/:service_id/preview/text-first')
```

  Adding the request.script_name as suggested here:
  https://github.com/rails/rails/issues/7977 makes the redirection
  to work as expected

Also using request.env['PATH_INFO'] (which is 'text-first'
works instead of request.path (which was
'/services/:service_id/preview/text-first')

I tried to add a current page as a hidden field but also has some
drawbacks so I rollback. If we don't add as hidden field everything breaks which
makes this approach more error-prone.

I also remove spring: This is not needed on the dummy app because we don't need t
o dummy app to be quick reloading as we don't often change the app, only
the gem

## Decisions

I am still not 100% happy with this approach but I think it is better than before.
Now the post request expects a page.URL which is the current page.

Example:

You are on the first page and submit the form:

        The app will make a POST /reserved/first-page/answers
        and then redirect to /second-page.

Then when you are on the second page and submit the form:

        The app will make a POST /reserved/second-page/answers
        and then redirect to the next page.

### DIFF
--- a/app/controllers/metadata_presenter/service_controller.rb
+++ b/app/controllers/metadata_presenter/service_controller.rb
@@ -2,32 +2,29 @@ class MetadataPresenter::ServiceController < MetadataPresenter.parent_controller
   helper MetadataPresenter::ApplicationHelper
 
   def start
-    @service = service
-    @start_page = service.start_page
-    render template: @start_page.template
+    @page = service.start_page
+    render template: @page.template
   end
 
   def answers
     save_user_data # method signature
 
-    # TODO: We need to see what to do with these line
-    # because of bots/hackers/etc
-    current_page = URI(request.referer).path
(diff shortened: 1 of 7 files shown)


brenetic:
we're currently not testing the `answers` method or the `render_page` method. if we're worried that it could be brittle we defo should test those

brenetic (approved):
🚀

PR #12, opened 2020-12-02
Add feature specs

tomas-stefano:
## Context

This PR adds feature specs to the gem.

We add capybara, simplecov and site prism gem (see commits for more details).

We add tests only for the single question form as that's what we currently support atm.

Co-authored-by: Natalie Seeto natalie.seeto@digital.justice.gov.uk


brenetic (approved):
🚀

njseeto (approved):

PR #13, opened 2020-12-03
Use govuk design system formbuilder gem

brenetic:
## Update govspeak to latest version

For some reason the presenter was a waaaaaaaay old version which had a vulnerability in kramdown. Use the lattest version. This will also be picked up by the fb-runner and will silence the notification there as well.

## Use govuk design system formbuilder gem

The old runner makes use of `govuk-frontend` built in macros to dynamically generate html pages on the fly. An example template file:

```
{%- macro text(data) %}
{% set params = setError(data) %}
{% set params = setLabel(params) %}
{% set params = setInputWidthClass(params) %}
{% if params.spellcheck %}
{% set attributes = setObjectProperty(params.attributes, 'spellcheck', params.spellcheck) %}
{% set params = setObjectProperty(params, 'attributes', attributes) %}
{% endif %}
{{ callMacro('govukInput', params) }}
{% endmacro -%}
```

This template, which is for a text field, is basically creating the necessary params which is then passed into the `govukInput` macro. That macro comes bundled with `govuk-frontend` npm package. It does a lot:

https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/components/input/template.njk

Nearly all the template files in fb-components makes calls to the  myriad of `govuk-frontend` helper macros.

This presents another quandary for us as nunjucks cannot be used directly inside rails.

DfE currently have a helpful gem which outputs govuk design system friendly components. This will help us at component level for sure. It does not support every single component that we require but it is open source so we could contribute to it or even fork it and extend it when we require it. For now it has the necessary components to not be blocking.

We will have to hand craft the page templates unfortunately.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,7 +2,8 @@ PATH
   remote: .
   specs:
     metadata_presenter (0.1.0)
-      govspeak
+      govspeak (>= 6.5.10)
+      govuk_design_system_formbuilder (>= 2.1.5)
       rails (~> 6.0.3, >= 6.0.3.4)
 
 GEM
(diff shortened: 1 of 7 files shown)


tomas-stefano (approved):
🚀 🎸

njseeto (approved):
🚀

PR #14, opened 2020-12-03
Application and head layouts

brenetic:
## Add page data block ID tags

The data-block-type for a page should be "page". The data-block-pagetype should be the type of the page that is being loaded. e.g singlequestion, form, start_page etc

## Use application and head layouts in presenter

By moving the main application layout in to the gem, as well as the header template, we are one step closer to the presenter gem being able to serve a fully compliant gov.uk page.

The assets are currently still expected to be in the containing app (editor and runner) but we can revisit this in the future.

To make use of this layout `layout: metadata_presenter/application` needs to be added to the `application_controller`.


njseeto (approved):
💯

PR #15, opened 2020-12-04
Default metadata for required fields

brenetic:
This adds the default metadata for a required field.

We are also caching the default metadata on startup.


tomas-stefano (approved):
🚀

njseeto (approved):

PR #16, opened 2020-12-07
Move schemas to presenter

brenetic:
## Add schemas to MetadataPresenter

The MetadataPresenter is fast becoming the most obvious place for holding things like the schemas and the default metadata that will be used by the fb-editor in the future. Having a single place for both of these allows us to make use of a single point for fixtures or perhaps a future generator.

This means that we will no longer have to juggle several different versions of metadata between the three apps that will be relying on them. The presenter is the single source of truth.

The one downside of this approach is that when the MetadataPresenter is added to the metadata-api, where schema validation occurs on the incoming requests, the latter will also have a copy of the views that it will never make use of.

## Validate schemas via the presenter

The MetadataPresenter will now also take care of schema validation. Any Form Builder app which installs the MetadataPresenter will now be able to validate any metadata it creates or receives in a consistent fashion.

## Added schemas for the start page and the single question page

Added schemas for the start page as well as the single question page.

Schema validation should now occur for page types as well.

## Add the service configuration schemas

The configuration for a given service takes the form of a service object, a meta object and an array of module objects (like save and return). The last of those we will not be moving in this PR.

Configuration is now an object that expects to have a service property and a meta property. The latter is the configuration for the service footer.

## Use Version fixture instead of Service fixture

The service fixture is the minimum metadata required for a creating a new service with the metadata-api. the version fixture is a fuller example of an actual form's metadata.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -4,6 +4,7 @@ PATH
     metadata_presenter (0.1.0)
       govspeak (>= 6.5.10)
       govuk_design_system_formbuilder (>= 2.1.5)
+      json-schema (>= 2.8.1)
       rails (~> 6.0.3, >= 6.0.3.4)
 
 GEM
@@ -106,7 +107,7 @@ GEM
       nokogumbo (~> 2)
       rinku (~> 2.0)
       sanitize (>= 5.2.1, < 6)
-    govuk_app_config (2.7.1)
+    govuk_app_config (2.8.0)
       logstasher (>= 1.2.2, < 1.4.0)
       sentry-raven (~> 3.1.1)
       statsd-ruby (~> 1.4.0)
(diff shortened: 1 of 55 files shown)


mattempty (approved):
🚀

njseeto (commented):
Nice job - thanks for doing this colossal task of moving the schemas!
One question, will we need to test the `app/validators/metadata_presenter/validate_schema.rb` here or is this tested elsewhere?

brenetic:
@njseeto you are absolutely right. this is step one. if you all agree with this approach, then my following PR is on the metadata-api to start using this gem, and to remove all the schemas there. the metadata-api is using the validation mechanism there for all the integration specs and also for testing the metadata serialiser.

i think we defo need to have unit tests for here for the default metadata that we move from the fb-components and we should add those as we migrate.

tomas-stefano (approved):
🚀 🎸

njseeto (approved):
💯 💯

PR #17, opened 2020-12-07
Use govuk_error_summary component

brenetic:
This should allow for the rendering of the errors summary list with links back to the individual element that triggered the error.

https://trello.com/c/Crwo6jqv/1124-validation-for-the-text-field-component

### DIFF
--- a/app/models/metadata_presenter/metadata.rb
+++ b/app/models/metadata_presenter/metadata.rb
@@ -2,10 +2,11 @@ class MetadataPresenter::Metadata
   include ActiveModel::Conversion
   extend ActiveModel::Naming
 
-  attr_reader :metadata
+  attr_reader :metadata, :errors
 
   def initialize(metadata)
     @metadata = OpenStruct.new(metadata)
+    @errors = ActiveModel::Errors.new(self)
   end
 
   def id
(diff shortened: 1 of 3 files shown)


app/models/metadata_presenter/metadata.rb
```diff
@@ -2,10 +2,11 @@ class MetadataPresenter::Metadata
   include ActiveModel::Conversion
   extend ActiveModel::Naming
 
-  attr_reader :metadata
+  attr_reader :metadata, :errors
 
   def initialize(metadata)
     @metadata = OpenStruct.new(metadata)
+    @errors = ActiveModel::Errors.new(self)
```
tomas-stefano, line 9: How this error is filled when he validation is triggered?

brenetic: currently it isn't coz we haven't written the `validate!` method yet 😆

tomas-stefano (approved):
🚀 🎸

app/models/metadata_presenter/metadata.rb
tomas-stefano: I see, I understand the slack message now 😅

tomas-stefano (approved):
🚀

PR #19, opened 2020-12-08
Add back links to pages

brenetic:
This adds the govuk design system back links to all pages except the start page.

In the future the confirmation page will probably be the only other page that doesn't need a back link.

The logic for previous page is super simple at this point, based entirely upon the order of the page objects in the service's pages array.

In the future we will have to cater for conditional logic for pages therefore finding the previous page will get a lot more complicated.

No acceptance tests were changed in the making of this PR. This is because they make use of the dummy app. Currently the assets which the govuk frontend expects reside in the containing apps; ie. fb-runner and fb-editor.

As a result of this in order for the navigation specs to pass with these changes we need to delegate the layout to the presenter engine, as happens in the fb-runner. In doing that all the asset tags then need to be installed in the dummy app. This would result in a massive amount of duplication that we would have to keep up to date.

We are currently considering where these acceptance should live going forward.

https://trello.com/c/KSu2E2bP/1089-back-links

### DIFF
--- a/app/models/metadata_presenter/service.rb
+++ b/app/models/metadata_presenter/service.rb
@@ -15,4 +15,8 @@ class MetadataPresenter::Service < MetadataPresenter::Metadata
     current_page = find_page(from)
     pages[pages.index(current_page) + 1] if current_page.present?
   end
+
+  def previous_page(current_page:)
+    pages[pages.index(current_page) - 1] unless current_page == start_page
+  end
 end
--- a/app/views/layouts/metadata_presenter/application.html.erb
+++ b/app/views/layouts/metadata_presenter/application.html.erb
@@ -21,6 +21,10 @@
     <%= render template: 'metadata_presenter/header/show' %>
     <div class="govuk-width-container govuk-body-m">
       <main class="govuk-main-wrapper govuk-main-wrapper--auto-spacing" id="main-content" role="main">
+      <% if @back_link %>
+        <a class="govuk-back-link" href="<%= @back_link %>">Back</a
+      <% end %>
+
         <%= yield %>
       </main>
     </div>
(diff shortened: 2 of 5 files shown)


app/models/metadata_presenter/service.rb
```diff
@@ -15,4 +15,8 @@ def next_page(from:)
     current_page = find_page(from)
     pages[pages.index(current_page) + 1] if current_page.present?
   end
+
+  def back_link(current_page: )
+    pages[pages.index(current_page) - 1]&.url
+  end
```
tomas-stefano, line 21: Because we have the `next_page`, can we call `previous_page` here?

app/views/layouts/metadata_presenter/application.html.erb
```diff
@@ -21,6 +21,10 @@
     <%= render template: 'metadata_presenter/header/show' %>
     <div class="govuk-width-container govuk-body-m">
       <main class="govuk-main-wrapper govuk-main-wrapper--auto-spacing" id="main-content" role="main">
+        <% if @back_link %>
+        <a class="govuk-back-link" href="<%= @back_link %>">Back</a
```
tomas-stefano, line 25: Small indentation

tomas-stefano (dismissed):
Minor questions but looks good! 🚀

app/views/layouts/metadata_presenter/application.html.erb
brenetic: I always never know which is best 😆 

if we indent the `a` then it's 4 spaces indented from `main` instead of 2. we could dedent the `<% if. . .`?

njseeto (approved):
🚀

app/views/layouts/metadata_presenter/application.html.erb
tomas-stefano: Me too 😆 Yes dedent the if sounds good!

tomas-stefano (approved):

tomas-stefano (approved):
🚀 🎸 ⭐

njseeto (approved):
🚀 🚀

PR #20, opened 2020-12-09
Remove feature specs

brenetic:
We are moving the features specs to the fb-runner, which was done in [this PR](https://github.com/ministryofjustice/fb-runner/pull/34).

The reason for this is to do with the future difficulty of keeping the dummy app in sync as it currently has no assets installed.


spec/spec_helper.rb
```diff
@@ -103,7 +103,7 @@
   add_filter '/spec/'
 end
 
-SimpleCov.minimum_coverage 97
+SimpleCov.minimum_coverage 80
```
njseeto, line 106: Out of curiosity, what's the reason for lowering this? Is 80% the coverage for all of our projects?

njseeto (approved):
Looks great - just left a comment about code coverage but otherwise is it all good! ⭐

spec/spec_helper.rb
brenetic: it's coz we're not doing any controller tests at the moment. we were relying on the features spec.

PR #24, opened 2020-12-16
Use Kramdown instead of Govspeak

brenetic:
This will hopefully be a temporary change. Govspeak has a dependency on govuk_publishing_components which is causing issues with the metdata-api at boot time in prod.

For now use Kramdown directly until we find a workaround.


tomas-stefano (approved):
🎸

njseeto (approved):
🚀

PR #27, opened 2020-12-18
Feature/change your answer

tomas-stefano:
## Context

This PR adds the possibility to change your answer from the check your answers page.

## Considerations on terminology

 We called "check your answer" but the metadata is defined as "page.summary". 

About terminology what should we use?
1. Check your answer
2. Page summary
3. Just summary
4. MetaSummary 3000

## Considerations on check your answers page

In order to know the page URL that the user should change we had to change the summary view to loop through each page and each component from each page. This can change when we start to add conditional logic.

## Common Considerations

Just renamed ApplicationController to EngineController to avoid any clashes with naming. 
Create a redirect_to_page because we found bugs on the editor due to the fact that `redirect_to '/some-url'` could clash with editor routes. The `redirect_to_page` uses the `script_name` to avoid those clashes.

### DIFF
--- a/app/controllers/metadata_presenter/application_controller.rb
+++ b/app/controllers/metadata_presenter/application_controller.rb
@@ -1,5 +1,0 @@
-module MetadataPresenter
-  class ApplicationController < ActionController::Base
-    protect_from_forgery with: :exception
-  end
-end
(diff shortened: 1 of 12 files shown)


njseeto (approved):
🚀

PR #28, opened 2020-12-18
Divide the service controller into smaller controllers

tomas-stefano:
Decided to make the service controller smaller.

Create

1. PagesController for rendering the templates
2. SubmissionsController for handling the submission (will be done
   in the future) and rendering the confirmation page
3. AnswersController for handling the user navigating through the form
4. ChangeAnswerController for handling the simple change answer logic


brenetic (approved):
🚀

PR #29, opened 2020-12-18
Docs/create

tomas-stefano:
## Context

This PR adds documentation to the project:

1. Readme
2. Yard as documentation generator for the gem.

I documented the base validator as example but since the whole gem has the public interface we should document more.
I also added the instructions in the Readme for generating the docs and to use the gem.


njseeto (dismissed):
🚀

brenetic (approved):
🚀

PR #31, opened 2020-12-23
Allow optional fields with additional validations

brenetic:
In scenarios where the question is optional and it is answered then additional validations should happen. If the question is not answered then additional validations should not happen.

The tests for this rely on the the `version.json` fixture `parent-name` page component being optional (required: false) but also having validations for minimum and maximum length. Therefore a user should be allowed to progress through the form without answering the question, however should they provide an answer, only then validate.

https://trello.com/c/XEuGMQnV/1156-bug-allow-optional-fields-with-other-validations


njseeto (dismissed):
🚀

njseeto (approved):

PR #35, opened 2020-12-30
Publish to Rubygems

brenetic:
## Add Brakeman gem 

For testing for obvious security issues. We are ignoring a couple of warnings at the moment. Documented in the `brakeman.ignore` file. One is to do with user answers which we cannot cater for specific keys as we do not know what the form owner may use. The second is for how we look for the next page currently. 

## Publish to Rubygems

We can start publishing versions to Rubygems going forward.

Remove the Gemfile.lock as gems don't need it.

Bump the version to 0.1.1, just coz . . .


njseeto (approved):
🚀

PR #36, opened 2020-12-30
Move fixtures to lib directory

brenetic:
When the gem is built the spec directory is not built with it, understandably. Therefore when the runner, editor and metadata-api look for the fixtures during startup or testing they cannot find them.

For the short term expose the fixtures directory using a method which can be called by the other apps. As the containing apps take shape they will no longer need fixtures at start up, only during testing. At that point we can make the method a spec helper of some kind.

Bump version to 0.1.2


njseeto (approved):
LGTM! 👍

PR #37, opened 2020-12-30
Add required additional directories to gem build

brenetic:
I completely forgot that the apps that use the built gem also require the `default_metadata` and `schemas` directories!

Since we are adding those additional directories we may as well add a `fixture` directory.

Add the fixture_directory method as a helper to for the specs instead.


njseeto (approved):
💯

PR #40, opened 2021-01-11
Do not require service_id in the base schema

brenetic:
Previously we separated out the service creation and version creation for a service into two separate schemas. This was due to the fact that we set service_id as a required property on the base.json schema.

However the metadata sets the service id on first creation and then after that the only way to create a new version is to send the service id as a path param. Therefore we only need to have a single request schema now instead of one for service creation and one for version creation.


tomas-stefano (approved):
🎸

PR #41, opened 2021-01-12
Update start page default metadata

brenetic:
This  just updates the content of the default metadata to match the example in the [Design System](https://design-system.service.gov.uk/patterns/start-pages/).

Also update the fixtures to match and remove some of the additional properties from the configuration that is no longer part of the default metadata.


tomas-stefano (approved):
🎸

PR #45, opened 2021-01-21
Check Answers

njseeto:
We have decided to change the Check Your Answers page schema from 'summary' to 'checkanswers'.
This aligns better with the GOVUK design systems: https://design-system.service.gov.uk/patterns/check-answers/

I have also added an empty array value for the components key in the default metadata as it is possible to add components into the check your answers page.

This change will help set the foundation for the story: https://trello.com/c/zp9OuUox/1196-backend-check-your-answers-page

This PR will need to be merged to allow the runner to work.


tomas-stefano (approved):
🎸

PR #47, opened 2021-01-25
Use UUID in page path

brenetic:
## Context

We address two major issues in the presenter:

1. Let's use the page uuid for base finder and not the url 
2. Let's generate answer routes for each url of the pages of the form

## Page uuid

We are generating a page uuid on the editor and use to find the pages in the editor.

## Generate answer routes

Before we had to add the page URL to the params on the runner so it would POST to /reserved-answers/%2f/ (%2f mean the '/'). We don't need to do that anymore because we are generating the POST to each URL of the pages metadata.

Example:

if we have a page with a url '/name' there will be a POST '/name' for that page on the presenter
Also return 404 if requested page URL does not exist in the metadata.

Release version 0.3.0

https://trello.com/c/wbZcslTs/1224-use-generated-page-id-instead-of-name

### DIFF
--- a/app/controllers/metadata_presenter/answers_controller.rb
+++ b/app/controllers/metadata_presenter/answers_controller.rb
@@ -1,5 +1,7 @@
 module MetadataPresenter
   class AnswersController < EngineController
+    before_action :check_page_exists
+
     def create
       if page.validate_answers(answers_params)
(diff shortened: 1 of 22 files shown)


njseeto (approved):
🚀

PR #53, opened 2021-02-08
Component/textarea

tomas-stefano:
## Context

This PR adds the textarea component to the metadata presenter gem.

The textarea supports:

1. Max words
2. Max chars
3. Threshold
4. Rows

The first three belongs to the character count feature:  https://design-system.service.gov.uk/components/character-count/

Another detail is that we should be making sure the validation is required by default in all components.


brenetic (approved):
🚀

PR #54, opened 2021-02-09
Only render pages which are not assets

tomas-stefano:
## Context

We need to fix the runner when the runner is published. Since we will make the runner serve the assets for now, we need to make sure that our global route match don't consider the assets as a "page" object.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -6,5 +6,7 @@ MetadataPresenter::Engine.routes.draw do
 
   post '/', to: 'answers#create'
   match '*path', to: 'answers#create', via: :post
-  match '*path', to: 'pages#show', via: :all
+  match '*path', to: 'pages#show',
+    via: :all,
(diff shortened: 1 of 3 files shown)


config/routes.rb
```diff
@@ -6,5 +6,7 @@
 
   post '/', to: 'answers#create'
   match '*path', to: 'answers#create', via: :post
-  match '*path', to: 'pages#show', via: :all
+  match '*path', to: 'pages#show',
+    via: :all,
+    constraints: lambda {|req| req.path !~ /\.(png|jpg|js|css|ico)$/ }
```
brenetic, line 11: this does make me cry some 😢

brenetic (approved):
🚀

config/routes.rb
tomas-stefano: When we serve via CDN this won't be needed.

PR #57, opened 2021-02-10
Component/number

tomas-stefano:
[Trello card](https://trello.com/c/LcEoCP68/1235-add-number-component-to-a-single-question-page)

## Context

This PR adds the number component to the gem.

The number component is very similar to the text component with the only gotcha that only accepts number (obviously 😅 ).

## Next steps

- [ ] Runner PR - fix specs because of the version.json fixture change
- [ ] Editor PR - bump the gem so the create single question -> number component to work 

🎸


brenetic (approved):
🚀

PR #59, opened 2021-02-10
Component/number

tomas-stefano:
## Context

I found a bug to reproduce.

Have on the metadata any validation **before** the required node:

```
  validation: {
      min_length: 10
      required: true
  }
```

It was making the min length to be triggering before the required and showing the min length validation when user answer is blank.

So I changed the logic to select all validations which are true and all of them should allow blank and only the required validator should validate the blanks.

This makes the code more reliable as any future validation that we will write will return valid if the user answer is blank letting the required validator to have the responsibilities to validate the empty inputs.

### DIFF
--- a/app/validators/metadata_presenter/base_validator.rb
+++ b/app/validators/metadata_presenter/base_validator.rb
@@ -42,6 +42,8 @@ module MetadataPresenter
     end
 
     def valid?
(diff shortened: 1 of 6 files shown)


njseeto (approved):
🚀

PR #61, opened 2021-02-12
Page answers object

tomas-stefano:
## Context

All the components needs to render the right values when user return to the page that the user already completed. The gov UK DfE convention is to call the value method on the form object to render the input value. So we need an intermediary object to re-render the page with the user answers already there.


njseeto (approved):
🚀

PR #62, opened 2021-02-12
Handle Radio Button components in the metadata

brenetic:
## Handle radio button components in metadata

This adds the ability for the presenter to handle the radio buttons component as well as validate it as a required field.

We added the humanised_title method in order to be able to correctly choose between using a label or legend property for page headings/questions, the check your answers page, as well as the key when creating the errors hash on the page answers object.

## Default radios metadata does not require label property

The radios metadata makes use of the legend as the main title or question for a given page.

## Use correct govuk heading css class for legend properties 

govuk-heading-xl is the correct css class required for the h1 of a page.

## Publish version 0.11.0

https://trello.com/c/vo4dqCu4/1237-add-radio-buttons-component-to-a-single-question-page


tomas-stefano (approved):
🎸

PR #63, opened 2021-02-12
Component/date

tomas-stefano:
## Context

This adds the date component to the gem.

## Next steps

- [ ] Runner acceptance tests
- [ ] Editor bump of the version

### DIFF
--- a/app/models/metadata_presenter/page_answers.rb
+++ b/app/models/metadata_presenter/page_answers.rb
@@ -1,4 +1,22 @@
 module MetadataPresenter
+  class DateField
+    attr_reader :day, :month, :year
+
+    def initialize(day:, month:, year:)
+      @day = day
+      @month = month
+      @year = year
+    end
+
+    def present?
+      !blank?
+    end
+
+    def blank?
+      @day.blank? || @month.blank? || @year.blank?
+    end
+  end
+
   class PageAnswers
     include ActiveModel::Model
     include ActiveModel::Validations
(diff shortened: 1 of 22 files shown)


app/models/metadata_presenter/page_answers.rb
```diff
@@ -17,11 +17,33 @@ def components
     end
 
     def respond_to_missing?(method_name, include_private = false)
-      answers[method_name.to_s].present?
+      method_name.to_s.in?(components.map(&:id))
```
brenetic, line 20: will this also need to cater for the three different date fields?

tomas-stefano: Yes because the gem calls the #try method which verifies if the method exists.
Since the `answers` for the date field is using the segments (3i, 2i, 1i) there is no answer that represents the id. Example:

Given we have a page 

```
   {
     #...
     components: [{
        id: 'birthday',
        type: 'date'
     }]
   }
```

The answers would be:

```
   { birthday(3i) => '31'. birthday(2i) => '12'. birthday(3i) => '2020' }
```

Then there is no birthday in the answer. Because we are using the id as the names we can infer by using the components id for the object. Otherwise, the date doesn't work (and it is tested in the runner acceptance tests. - future PR that will be opened)

tomas-stefano: Sorry, now I read the question. This line does the trick and there is no additional work to be done.

brenetic (approved):

PR #67, opened 2021-02-16
Component/checkbox

njseeto:
This PR:
- adds items to the checkbox default metadata, which is required for the editor
- Removes `name` from the required fields as it is not used
- Bumps the version to 0.13.2


brenetic (approved):
🚀

PR #69, opened 2021-02-17
Suppress hidden fields in checkbox collections

njseeto:
Now that the dfe formbuilder gem allows `include_hidden: false`, we can use this in our checkbox collection and remove our old workaround.

This PR also bumps the metadata-presenter version to 0.13.4.

For context: https://github.com/DFE-Digital/govuk_design_system_formbuilder/pull/240

### DIFF
--- a/app/validators/metadata_presenter/base_validator.rb
+++ b/app/validators/metadata_presenter/base_validator.rb
@@ -63,12 +63,7 @@ module MetadataPresenter
     # @return [String] user answer for the specific component
     #
     def user_answer
-      value = page_answers.send(component.name)
-      if component.type == 'checkboxes'
-        Array(value).reject(&:blank?)
-      else
-        value
-      end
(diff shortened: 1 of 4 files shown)


app/validators/metadata_presenter/base_validator.rb
```diff
@@ -65,7 +65,7 @@ def custom_error_message
     def user_answer
       value = page_answers.send(component.name)
       if component.type == 'checkboxes'
-        Array(value).reject(&:blank?)
+        Array(value)
```
brenetic, line 68: i have a hunch that this method _might_ be able to be returned to `page_answers.send(component.name)`

njseeto: ah yes you are correct

brenetic (approved):
🚀

PR #80, opened 2021-03-04
Multiple questions page

brenetic:
## Add multiple question schemas and default metadata

This adds the multiple questions schema and metadata.

It was necessary to adjust how the PageAnswersPresenter collected the answers when passing them into the CYA page.

Previously the map function returned a flat array of Presenters. When there is only one question per page this is fine as the CYA page uses the label or the legend of the component and does not think about the page at all.

However for multiple question pages the questions are essentially grouped by the heading on the page. By the time we get to the CYA page it makes sense to group the answers by the page heading and present that back to the user.

In order to achieve this we now collect the page answers grouped by the page that they are found on, then iterate over those pages and then the answers checking whether it is necessary to show the heading of the page or not.

## Multiple Questions Page

Screenshot 2021-03-04 at 14 56 56

## Check Your Answers Page (After)

Screenshot 2021-03-04 at 11 22 09

https://trello.com/c/ZiXnSx8Q/1335-multi-question-page

### DIFF
--- a/app/views/metadata_presenter/page/checkanswers.html.erb
+++ b/app/views/metadata_presenter/page/checkanswers.html.erb
@@ -35,25 +35,36 @@
     <%= form_for @page, url: reserved_submissions_path do |f| %>
       <div data-block-id="page.checkanswers.answers" data-block-type="answers">
         <dl class="fb-block fb-block-answers govuk-summary-list">
-          <% page_answers_presenters.each do |page_answers_presenter| %>
-            <div class="govuk-summary-list__row">
-              <dt class="govuk-summary-list__key">
-                <%= page_answers_presenter.humanised_title %>
-              </dt>
+          <% pages_presenters.each do |page_answers_presenters| %>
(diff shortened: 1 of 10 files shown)


app/views/metadata_presenter/page/checkanswers.html.erb
```diff
@@ -35,25 +35,36 @@
...
-              <%= link_to(
-                    change_answer_path(url: page_answers_presenter.url),
-                    class: 'govuk-link'
-                  ) do %>
-                     Change<span class="govuk-visually-hidden"> Your answer for <%= page_answers_presenter.humanised_title %></span>
-                <% end %>
-                </a>
-              </dd>
-            </div>
+              <% if page_answers_presenter.multiple_questions? && index == 0 %>
```
tomas-stefano, line 41: Can we create and move this condition to page_answers_presenter.display_heading?

tomas-stefano (approved):
:shipit:

njseeto (dismissed):
💯

tomas-stefano (approved):
🚢

PR #85, opened 2021-03-15
Add content page schema, default metadata and template

brenetic:
## Add content page schema and default metadata

This adds the content page schema, default metadata and the template. There is only meant to be the section heading, heading, lede and body. So very similar to the start page.

On top of that there will be the ability to add another content component in a future story.

The continue button remains as the page is meant to be for informative purposes. Since the public user is not going to be entering any information this content page will never appear on the check your answers page nor in the submission itself.

## Add the necessary class to the footer

This will make it sticky when used in the runner

## Use correct heading class on start page

The page object has now heading_class method. Set it to be govuk-heading-xl for now. This makes it match all the other page headings.

## 0.18.0

https://trello.com/c/vnNbK8rZ/1370-content-page-mvp-version-m-add-and-edit


tomas-stefano (approved):
🚀

PR #89, opened 2021-03-16
Do not parse raw HTML in certain element fields

brenetic:
We only want markdown to be parsed in order to correctly display formatted html back to the user. Therefore we do not want to support an editor entering raw html into a section heading, heading, question or lede.


tomas-stefano (approved):

njseeto (dismissed):

mattempty (approved):
🚀

njseeto (approved):

PR #90, opened 2021-03-18
Default text for elements and components

njseeto:
We set the initializer method for the metadata object to now take an attribute called editor which by default will be false. This is used to check whether or not to show any default text for any given element.

In the editor we want default text to be shown if the user has not entered any custom text into a field. However in the runner if the editor user has not entered any custom text then we do not want the default text to be shown to the public user.

Co-authored-by: Brendan Butler <brendan.butler@digital.justice.gov.uk>

Story: https://trello.com/c/YSHgwzER/1382-placeholder-text


tomas-stefano (approved):

brenetic (approved):
🚀

PR #91, opened 2021-03-18
Add id to item model

tomas-stefano:
Without this the radio/checkboxes will use the component.id for
the values in the radios and that's not what we want.


njseeto (approved):
🚀

PR #101, opened 2021-03-31
Use item id in collection components

brenetic:
## Use item ID in collection component templates

If we use the label then the DFE gem just generates all the collection items with the same IDs which then conflict.

Using label means that that gem would generate:

`answers-label-name-option-field`

But not increment any number to it, so would not be unique within a given collection.

As we generate item IDs each time we create a new collection component, and those IDs are always unique, we should use this.

We don't need to override the id method as the parent metadata object has it and is correctly retrieving _id from the passed in metadata JSON which in the instance of an Item and is the specific collection item, not the parent collection component.

Unfortunately this does mean that the generated item IDs will be a little silly. E.g

Given the page url 'some-page' and the component id 'some-component' the gem will now generate:

```
answers-some-page-some-component-1-some-page-some-component-item-1-field
answers-some-page-some-component-1-some-page-some-component-item-2-field
```

etc

## 0.25.0

https://trello.com/c/sv8f7IM1/1427-bug-radio-buttons-and-check-box-clicking-second-item-label

### DIFF
--- a/app/models/metadata_presenter/item.rb
+++ b/app/models/metadata_presenter/item.rb
@@ -1,8 +1,4 @@
 class MetadataPresenter::Item < MetadataPresenter::Metadata
-  def id
-    label
-  end
-
   def name
     label
   end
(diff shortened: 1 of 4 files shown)


tomas-stefano (approved):
🚀

PR #102, opened 2021-04-01
Revert Item ID back to being label

brenetic:
Revert Item ID back to being label

In reality an editor is not going to leave the label of a collection item to be the one set in the default metadata. The confusion comes from the fact that the label for an item in the default metadata is "Option" so when a user previews a form the collections do not work properly as the DFE gem uses the label to create the item ID. However its expectation is that the label will be unique, which is entirely reasonable.

So swap the Item ID back to being the label. Our expectation is that an editor will always change the labels to be unique.

Publish 0.26.0


njseeto (approved):
🚀

PR #108, opened 2021-04-09
Feature/non flow pages

brenetic:
## Make meta items available in the footer

These are links to standalone pages that should reside in the footer.

## Add standalone page schemas, default metadata and templates

This adds the concept of a standalone page. Since they are not part of the standard flow we did not want to add them to the already existing pages array. We may want to revisit this decision but for now this is fine.

By their very definition a standalone page exists independently of form.

## Do not use back link on confirmation page

The confirmation page does not require a back link as by this point the user has already submitted their form.

## Fall back to the referrer for previous page link

Pages like standalone pages are not part of our current implementation of flow pages. Therefore the previous_page method will never find the index of the previous page.

In these instances use the referrer URL.

## Validate the standalone pages on start up

We also want to validate that the standalone pages in the metadata are valid so iterate over those too.

## Screen shots

Screenshot 2021-04-09 at 15 27 16

Screenshot 2021-04-09 at 15 27 35

Screenshot 2021-04-09 at 15 27 57

Screenshot 2021-04-09 at 15 28 17

https://trello.com/c/LtJSIu15/1422-represent-non-flow-standalone-pages-for-mvp


app/models/metadata_presenter/service.rb
```diff
@@ -36,8 +44,29 @@ def confirmation_page
...
 
+  def meta
+    MetadataPresenter::Meta.new(configuration['meta'])
+  end
+
   private
 
+  def all_pages
+    @all_pages ||= pages + standalone_pages
+  end
```
tomas-stefano, line 55: So note to ourselves: more urls we need to restrict later (all meta pages plus the restricted ones)

spec/models/meta_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+            "text": 'Accessibility'
+          }
+        ]
+      }
+    end
+
+    it 'returns the an array of meta items' do
+      meta.items.each do |item|
+        expect(item).to be_kind_of(MetadataPresenter::MetaItem)
+      end
```
tomas-stefano, line 35: Do we need to verify the items size? `expect(meta.items.size).to be 3`. Just asking because if the implementation returns an empty array this test will pass.

spec/models/service_spec.rb
```diff
@@ -124,23 +124,52 @@
...
+        expect(previous_page.id).to eq(service.start_page.id)
+      end
+    end
+
+    context 'when there is no referrer page' do
+      it 'returns nil' do
+        previous_page = service.previous_page(current_page: nil, referrer: nil)
+        expect(previous_page).to be_nil
+      end
+    end
```
tomas-stefano, line 161: Do we need to include a test for the confirmation page?

app/models/metadata_presenter/service.rb
```diff
@@ -14,20 +20,22 @@ def service_slug
...
 
   def find_page_by_url(url)
-    pages.find { |page| strip_slash(page.url) == strip_slash(url) }
+    all_pages.find { |page| strip_slash(page.url) == strip_slash(url) }
   end
 
   def find_page_by_uuid(uuid)
-    pages.find { |page| page.uuid == uuid }
+    all_pages.find { |page| page.uuid == uuid }
   end
```
tomas-stefano, line 28: Do we need to add unit tests for `#find_page_by_url` and `find_page_by_uuid` testing that it can find the standalone pages?

tomas-stefano (approved):
Awesome work. I added minor questions but most of, are just questions so feel free to answer my questions/merge 🚀

app/models/metadata_presenter/service.rb
brenetic: have added it to the card for doing that work

tomas-stefano (approved):

PR #110, opened 2021-04-12
Fix/version issues

stevenburnell-moj:
Attempt to fix version issues where previous change not showing as result of jumping from 0.27.3 -> 0.28.0 due to conflict file resolution action. This commit essentially reintroducing what happened (originally meant for v0.27.4 in between).


stevenburnell-moj:
Closing before it gets approved since there are more things to fix and I can add to this.

PR #111, opened 2021-04-12
Fix/version issues

stevenburnell-moj:
- Version confusion missed HTML attribute change for ID
- Confirmation panel had smaller black text (should be white)


brenetic (approved):
🚀

PR #120, opened 2021-04-16
F marry patch fields

f-marry:
Attempting to remove the unwanted section heading on CYA page, and remove the unwanted text on the start page.


tomas-stefano (approved):
🚀

PR #121, opened 2021-04-16
Fix Check answers table layout

brenetic:
## Fix CYA table layout to not include content components

The CYA table was accidentally including content components in the layout. These should be removed from the components that the answers presenter gathers.

Also if there was a multiple question we need to close the </dl> tag and open another <dl> tag when it's the last question for that page. Otherwise it groups the next question into the same part of the table as the multiple questions pages questions.

## Before

Screenshot 2021-04-16 at 19 12 06

## After

Screenshot 2021-04-16 at 19 12 27

### DIFF
--- a/app/models/metadata_presenter/page.rb
+++ b/app/models/metadata_presenter/page.rb
@@ -22,12 +22,24 @@ module MetadataPresenter
       to_h.reject { |k, _| k.in?(NOT_EDITABLE) }
     end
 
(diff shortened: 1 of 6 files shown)


tomas-stefano (dismissed):
🚀  Plus, can you write feature specs in the runner repo for this PR please?

njseeto (approved):
🚀

PR #125, opened 2021-04-22
Update Cookies page

njseeto:
Update the cookies page as per content design changes.
This is because we are having difficulty parsing Markdown tables

Story: https://trello.com/c/ZipHL83H/1493-update-the-standalone-pages

Note: I have not bumped the version, please let me know if this should be bumped to `0.1.1`

Once approved the Cookies page will look like this:

Screenshot 2021-04-22 at 08 18 43


mattempty (approved):
🚀

PR #128, opened 2021-04-27
Changes the label for added radio option from ["Required option text"] to "option" 

f-marry:
Changes the label for added radio option from ["Required option text"] to "option" so the new option and the "existing options" will have the same label as they should.


brenetic (approved):
🚀

PR #133, opened 2021-05-04
Footer section break

brenetic:
Remove the section break in the footer

Publish 1.0.7

## Before

Screenshot 2021-05-04 at 17 09 23

## After

Screenshot 2021-05-04 at 17 11 37


njseeto (approved):
🚀

PR #138, opened 2021-05-24
Feature/upload files

tomas-stefano:
## Context

This PR adds the upload mechanism to the gem and its current validators 🚀 

Validations:

* File type
* File max size
* File virus scan

## Upload mechanism

Basically, it calls the upload_file method implemented on the mounted app if there is a upload component on the current page.


njseeto (approved):
🚀

PR #143, opened 2021-06-04
Feature/move upload to gem

tomas-stefano:
## Context

Make the code more cohesive by moving the upload logic to the gem


brenetic (approved):
🚀

PR #144, opened 2021-06-04
Return empty hash when answer is blank

tomas-stefano:
When we are using filestore and there is no file we shouldn't
call filestore and we should return an empty uploaded file.

Co-authored-by: Brendan Butler <brendan.butler@digital.justice.gov.uk>


brenetic (approved):
🚀

PR #146, opened 2021-06-24
Branching v1

tomas-stefano:
[Trello card](https://trello.com/c/9rPzZHvX/1629-branching-evaluate-radio-components)

## Context

This is the simple branching version with radio buttons.

This approach only works for single criteria in many conditions.

The following operators are included:

1. Is
2. is not
3. is answered
4. is not answered

To create more operators you need to create a class with the same name as the operator that responds to #evaluate method. Example if we have the operator "my_awesome":

```ruby
module MetadataPresenter
  class MyAwesomeOperator < BaseOperator
     def evaluate?
        expected == "Awesome"
     end
  end
end
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ All notable changes to this project will be documented in this file.
 
 ### Changed
 
+## [1.4.0] - 2021-06-24
+
+### Added
+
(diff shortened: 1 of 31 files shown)


brenetic (approved):
🚀

PR #147, opened 2021-06-28
Branching checkboxes

tomas-stefano:
## Context

This PR adds support for checkboxes on evaluating conditions (a.k.a branching).

We add a method signature called `#evaluate_collection?` for checkboxes. If an operator needs special treatment for collections then just overwrite the method in the operator subclass.

## Rake script to humanize the flow

There is a rake script to humanize the flow but it is not 100% right yet. Please do let me know if you want to exclude it from the PR or leave it there to finish properly later.

### DIFF
--- a/spec/models/evaluate_conditions_spec.rb
+++ b/spec/models/evaluate_conditions_spec.rb
@@ -76,14 +76,53 @@ RSpec.describe MetadataPresenter::EvaluateConditions do
           end
         end
       end
+
+      context 'when the question is a checkbox component' do
+        let(:flow) { service.flow('618b7537-b42b-4551-ae7d-053afa4d9ca9') }
+
+        context 'when beef cheese and tomato condition "is" met' do
+          let(:user_data) do
+            {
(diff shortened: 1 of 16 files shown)


spec/models/evaluate_conditions_spec.rb
```diff
@@ -76,14 +76,55 @@
...
+        it 'returns the page uuid for the default page' do
+          expect(page).to eq(service.find_page_by_url('favourite-band'))
+        end
+      end
+
+      context 'for multiple answer (checkboxes) questions' do
+        let(:flow) { service.flow('ffadeb22-063b-4e4f-9502-bd753c706b1d') }
+        let(:user_data) { {} }
+        # this is probably more what the backend should be receiving
+        # let(:user_data) { { 'burgers_checkboxes_1' => [] } }
```
brenetic, line 123: `{ 'burgers_checkboxes_1' => [] }` is what the backend should be getting

tomas-stefano: ♻️

brenetic (approved):
🚀

PR #148, opened 2021-06-29
Generate flow diagrams

tomas-stefano:
## Context

This PR adds a rake task that generates a diagram for any metadata which contains the flow object.

## Running the rake task

It is added to the readme. You need to install Graphviz: 

`brew install graphviz`

Then you can run with `rails app:metadata:flow` (inside of the gem).

In the mounted app (a.k.a runner) you can run only `rails metadata:flow`.

### Running in the runner

It also accepts the `SERVICE_METADATA` env var:

```
SERVICE_METADATA="my awesome metadata"  rails metadata:flow
```

## Results

branching-fixture

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,3 +83,15 @@ service.
 ## Generate documentation
 
 Run `rake doc` and open the doc/index.html
+
+## Flow diagrams
+
+You can generate flow diagrams calling a rake task:
+
+```
+  brew install graphviz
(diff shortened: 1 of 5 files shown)


mattempty (commented):
Awesome. 🚀

brenetic (approved):
🚀

njseeto (approved):
🚀

PR #149, opened 2021-07-01
Branching previous page

tomas-stefano:
[Trello card](https://trello.com/c/95zZQwKB/1642-epic-change-backlinks-behaviour-in-the-form)

## Context

This PR adds the backlink feature for navigating on branches.

It evaluates all pages that the user traversed into. This could be used in the check answers page too.

### DIFF
--- a/app/models/metadata_presenter/previous_page.rb
+++ b/app/models/metadata_presenter/previous_page.rb
@@ -1,0 +1,27 @@
+module MetadataPresenter
+  class PreviousPage
+    include ActiveModel::Model
+    attr_accessor :service, :user_data, :current_page, :referer
+
+    def page
+      return if current_page.blank?
+
+      if flow.present?
+        previous_page = TraversedPages.new(service, user_data).last
+
+        return previous_page if previous_page
+
+        service.start_page
+      else
+        service.previous_page(
+          current_page: current_page,
+          referrer: referer
+        )
+      end
+    end
+
+    def flow
+      service.metadata['flow']
+    end
+  end
+end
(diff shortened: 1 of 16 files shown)


app/models/metadata_presenter/previous_page.rb
```diff
@@ -0,0 +1,38 @@
+module MetadataPresenter
+  class PreviousPage
+    include ActiveModel::Model
+    attr_accessor :service, :user_data, :current_page, :referrer
+
+    def page
+      # what happens when a user enters in the middle of the flow
```
brenetic, line 7: can we add a test for this?

tomas-stefano: Do you mean if the user jump pages? Then what should happen? Can you provide me with an example?

brenetic: user visits the start page, therefore there is a session. then they jump to the middle of a flow just because they know the URL for some reason.

should they go back to the start page as there are no answers? or should return to the previous page in the in the flow? Or just return to the referrer? am not sure what the correct behaviour should be

tomas-stefano: Cool, that's a good question. I'll ask the team!

tomas-stefano: Discussion on https://mojdt.slack.com/archives/CE78LKCU8/p1625144555035600
The conclusion is that we will create a specific card to handle the user jumping pages.

brenetic (approved):
🚀

PR #150, opened 2021-07-01
Don't require ruby graphviz in runtime, only when run rake task

tomas-stefano:
## Context 

Needs to fix the require ruby graphviz because the runner/editor/metadata-api doesn't require it.

The rake task will run only locally.


brenetic (approved):
🚀

brenetic (approved):
🚀

PR #151, opened 2021-07-05
Add support for conditions with "OR" and "AND"

tomas-stefano:
## Context

So all conditions can be mapped through criteria using "OR" and "AND".

So we can have a condition object with many criteria object and having the condition_type managing the page to be evaluated.

### DIFF
--- a/app/models/metadata_presenter/evaluate_conditions.rb
+++ b/app/models/metadata_presenter/evaluate_conditions.rb
@@ -4,24 +4,32 @@ module MetadataPresenter
     attr_accessor :service, :flow, :user_data
 
     def page
-      results = conditions.map do |condition|
-        condition.criterias.map do |criteria|
+      evaluated_page_uuid = page_uuid || flow.default_next
+
+      service.find_page_by_uuid(evaluated_page_uuid)
+    end
+
+    def page_uuid
+      @results ||= conditions.map do |condition|
+        evaluated_criterias = condition.criterias.map do |criteria|
           criteria.service = service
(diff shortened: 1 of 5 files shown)


brenetic (dismissed):
🚀

app/models/metadata_presenter/evaluate_conditions.rb
```diff
@@ -3,25 +3,35 @@ class EvaluateConditions
     include ActiveModel::Model
     attr_accessor :service, :flow, :user_data
 
+    AND_CONDITIONS = %w[IF AND].freeze
```
brenetic, line 6: am ok with these being lower case as well. save having to upcase the condition_type further down.

i'm not sure 'if' counts as an 'and' condition.

tomas-stefano: Maybe I should rename this constant to something else. In my view, we should use "if" only when we have a single criteria for the condition.

brenetic: i guess it depend what this will look like with the OR. we could go without the constant for now until we get to the OR?

tomas-stefano: Ok, so I will change the if to the other way around to use the OR instead.

tomas-stefano: Changed to use the other way around ♻️

brenetic (approved):
🚀

PR #152, opened 2021-07-06
Prevent code from being saved in the database

njseeto:
To mitigate against executable JavaScript being submitted, we have sanitised any HMTL tags that might used by a user.

Story: https://trello.com/c/oLpW8JdV/1599-confirm-code-snippets-are-not-saved-into-the-metadata


tomas-stefano (dismissed):
🚀

app/models/metadata_presenter/page_answers.rb
```diff
@@ -43,7 +44,7 @@ def upload_answer(component_id)
         file_details
```
brenetic, line 44: we still need to sanitize the original_filename inside this object

spec/models/page_answers_spec.rb
```diff
@@ -120,6 +120,109 @@
...
+          let(:answers) { { 'name_text_1' => 'script/' } }
+
+          it 'returns the value of the answer' do
+            expect(page_answers.name_text_1).to eq('script/')
+          end
+        end
+
+        context 'when the component is a textarea' do
+          let(:page) { service.find_page_by_url('family-hobbies') }
+          let(:answers) { { 'hobbies_textarea_1' => 'script/' } }
```
brenetic, line 136: can we make this part of a paragraph please. it's probably all fine but just wanna make sure it's not gonna rip out a random script word

tomas-stefano (approved):

PR #153, opened 2021-07-06
Flow object schemas

brenetic:
This adds the schemas for the two different types of flow objects:

- Page
- Branch

Internally in those objects there are also the condition and criteria
objects as well which deal with conditional logic.

Also add the default metadata for the following parts of flow objects:

- page
- branch
- condition
- criteria

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ All notable changes to this project will be documented in this file.
 
 ### Changed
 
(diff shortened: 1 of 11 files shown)


tomas-stefano (dismissed):
🎸

njseeto (dismissed):
🚀

njseeto (approved):

tomas-stefano (approved):

PR #156, opened 2021-07-07
Get all service flow objects

brenetic:
The previous flow method expected a uuid as an argument. It returns a
flow object which matches the provided uuid. It is not always a page
uuid but could also be a branch uuid so rename it to just uuid.

calling `flow` on a service should return all of the flow objects so
rename the current method to `flow_object` as that returns a single
object. This should allow the current implementation to simply return
all the flow objects for a service should they exist.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ All notable changes to this project will be documented in this file.
 
 ### Changed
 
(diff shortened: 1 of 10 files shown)


njseeto (approved):
🚀

tomas-stefano (approved):
:shipit:

PR #157, opened 2021-07-08
Create base flow schema

brenetic:
The service flow object uses UUIDs as the string property for either a
page object or a branch object. These UUIDs are obviously dynamic so
this schema has to allow for that as well.

Adding a base schema which makes use of the `patternProperties` to allow
for dynamic uuid string keys.

Some metadata objects do not have an '_id' property. In those instances
fallback to using the '_type' property when loading the metadata on
startup.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,17 @@ All notable changes to this project will be documented in this file.
 
 ### Changed
 
+## [1.7.5] - 2021-07-08
+
+### Added
+
+- Base flow object schema to cater for dynamic UUID properties
+
+### Changed
+
+- Remove '_id' property from flow objects default metadata
+- Fallback to using '_type' as the key for default metadata that does not have a '_id' property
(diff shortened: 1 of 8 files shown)


tomas-stefano (dismissed):

CHANGELOG.md
```diff
@@ -8,6 +8,12 @@ All notable changes to this project will be documented in this file.
 
 ### Changed
 
+## [1.7.4.1] - 2021-07-08
+
+### Added
+
+- Base flow object schema to cater for dyamic UUID properties
```
njseeto, line 15: small typo on dynamic

njseeto:
Small typo but otherwise looks good.
Happy for it to be merged once a decision has been made re: any string regex

brenetic:
@njseeto good eyes! thanks, have updated

njseeto (approved):
🚀

PR #159, opened 2021-07-09
Fixtures with flows

brenetic:
## Add flow object to default service metadata

All new services should have a flow object

## Add service and version fixtures with flow objects

These are direct copies of the current service and version fixtures but with additional flow objects. 

Keep the old ones as well so as not to break other tests currently. Eventually all tests should use the fixtures which include service flow objects

## Remove steps from start page default metadata

The Steps array is not used so we do not need to create it at all

## Publish 1.8.1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ All notable changes to this project will be documented in this file.
 
 ### Changed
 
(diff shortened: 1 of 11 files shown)


tomas-stefano (approved):
🚢

njseeto (approved):
:shipit:

PR #160, opened 2021-07-12
Branching submission

tomas-stefano:
## Context

On submission, we don't have the current page so we need to use the last page instead.

This PR adds this possibility.


brenetic (approved):
🚀

PR #161, opened 2021-07-13
Use Conditionals and Expressions

brenetic:
An unfortunately large change. We are attempting to correctly name these
properties to fulfill a little bit of a contentious subject.

Currently the design on the Editor uses the terms "Rules" and "Conditions".
"Rules" are really what developers think of as `conditionals`. i.e evaluate
"if", "and", "or" etc.

"Conditions", as they are used in the design, are actually what developers
think of as `expressions`. i.e evaluate "equal to", "greater than" etc.

While we know that there may be some confusion when it comes to mapping
these over in the backend we feel it is the correct terminology going
forward. If the design changes at some point to use whatever terminology
the end user finds the easiest to understand then we can just change the
mapping and not the schemas and metadata.

## Publish 2.0.0


tomas-stefano (approved):
🚀 🚀 🚀 🌔 🌞 🌌

PR #163, opened 2021-07-26
Only use service flow

brenetic:
## Only cater for services with flow objects

Now that the services have been migrated we can exclusively use the
service flow in order to determine the next page.

Update all fixtures to include a service flow.

## Back links will default to using the previous flow page

This may or may not end up being a temporary behaviour. When a user
is interacting with the service flow the back link works as intended.
However if they were to move onto a standalone page and then go back
into the service flow it is possible for them to get stuck in a loop
when clicking 'Back'.

Not a 100% ideal solution, but for the moment as soon as the user
goes back to a flow page then the back link will beging to extrapolate
the back link from the service flow. This is different to how a
browser may remember back as it holds state with all the users visited
pages which currently the runner does not.

It is possible that we should hold state for where the user has visited
if we want to mimic the behaviour of the browser.


tomas-stefano (approved):

njseeto (approved):
🚀

PR #164, opened 2021-07-27
Remove standalone pages backlinks

brenetic:
We have decided to remove back links entirely from standalone pages.
Only pages in the service flow will have a back link from now on.

This does mean that we can also remove a fair amount of code and related
tests as well.


njseeto (approved):
🚀

PR #165, opened 2021-07-30
Return a page object containing a given component

brenetic:
There are times when we need to get access to a page object which
has a given component.

Add a method to the service model which receives a component uuid and
returns the page object that contains it.

Co-authored-by Natalie Seeto <natalie.seeto@digital.justice.gov.uk>


tomas-stefano (approved):
🚀

PR #166, opened 2021-08-05
Supported components, input components and content components

brenetic:
## Supported components

The previous input_components and content_components methods did not
return actual component objects but rather just an array of the strings
relating to the supported components for a given page.

Change the name of those methods to better reflect their actual
function.

## Check if page is a question page 

There are some needs for only getting pages which are actually a
singlequestion or mulitiplequestions page so add a method to do that.

## Input and content components

This re creates the input_comonents and content_component methods to
actually return component objects of those types

## 2.2.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,18 @@ All notable changes to this project will be documented in this file.
 
 ### Changed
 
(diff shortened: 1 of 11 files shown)


njseeto (approved):
🚀

PR #168, opened 2021-08-09
Add branching title

tomas-stefano:
## Context

Return all branches from a service.

Used on https://github.com/ministryofjustice/fb-editor/pull/629


brenetic (approved):
🚀

PR #169, opened 2021-08-16
Surface flow object UUID

brenetic:
## Surface flow object uuid

The MetadataPresenter::Flow object has no knowledge of the page uuid
that it is related to. This can be a useful thing to surface on the flow
object.

Initialize the flow object with the uuid which is used as the key for
the object in the service flow hash.

## Return service flow as Flow objects

This adds a method which may prove helpful when working on the future
pages flow.

Previously we were only returning branches as MetadataPresenter::Flow
objects. `flow_objects` returns all flow objects, including branches.

## Publish 2.3.1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,17 @@ All notable changes to this project will be documented in this file.
 
 ### Changed
 
-## [2.2.0] - 2021-08-05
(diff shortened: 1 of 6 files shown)


tomas-stefano (approved):
💯

njseeto (approved):
🚀

PR #170, opened 2021-08-18
Remove lede

njseeto:
### Remove lede from the start page
The lede is not used in the start page so we have removed it from the `start.html.erb`, the default metadata, fixtures and the start page schema

### Bump to 2.3.2

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ All notable changes to this project will be documented in this file.
 
 ### Changed
 
+## [2.3.2] - 2021-08-18
+
+### Changed
+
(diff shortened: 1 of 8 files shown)


brenetic:
coolio. do the schema validations pass in the editor and the metadata api if they're using this version of the presenter?

njseeto:
Good point! Have tested on both the editor and the metadata api and they seem to pass in both 🎉

njseeto:
Will bump to version 2.3.2 on the editor and the metadata-api once this has been merged in

tomas-stefano (approved):
🎸

PR #172, opened 2021-08-24
Expression memoization

brenetic:
This is a small performance related change. There is no need to go and
keep looking up the Page object and the Component object once they have
been found previously.

There are times when we make several calls to find the item and label
for a given expression so this should save a few micro seconds of lookup
time.


mattempty (approved):
🚀

tomas-stefano (approved):

njseeto (approved):
🚀

PR #175, opened 2021-09-07
Change your answers

tomas-stefano:
[Trello card](https://trello.com/c/BmH2CAqX/1868-updated-branch-flow-cya-logic-for-branching-specific-questions)

## Change answers

This PR adds the change your answer feature.

## Explanations of what it was done

Regular question page: changing an answer on a non-branching page will jump the user back to the CYA page when they submit their updated answer.

Branching question page: changing a branching question will require the user to follow a new branch, enter information for the branch-specific question(s) and then back through the journey with all questions previously answered pre-populated for them to click through. 

Returning to a previously completed branch:  if a user then jumps back to a previous branch their answers are remembered. They can update as desired and then they traverse back through the form with all previous answers pre-populated but they are required to click back through the journey. 

Check Your Answers page: a user should only see the data related to their most recent journey e.g. if they’ve changed their answers 3x only the answers pertaining to Journey 3 should be displayed. 

Returning to a branching question page and keeping the same answer: if a user keeps the same answer on a page with branching then they should go straight back to the Check Your Answers page.

### DIFF
--- a/app/models/metadata_presenter/next_page.rb
+++ b/app/models/metadata_presenter/next_page.rb
@@ -1,7 +1,7 @@
 module MetadataPresenter
   class NextPage
     include ActiveModel::Model
-    attr_accessor :service, :session, :user_data, :current_page_url
+    attr_accessor :service, :session, :user_data, :current_page_url, :previous_answers
 
     def find
       return check_answers_page if return_to_check_your_answer?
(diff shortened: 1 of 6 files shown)


app/models/metadata_presenter/next_page.rb
```diff
@@ -11,17 +11,53 @@ def find
       else
         service.find_page_by_uuid(current_page_flow.default_next)
       end
+    ensure
+      session[:return_to_check_your_answer] = nil
```
brenetic, line 15: 👍  nice. was gonna ask about this

app/models/metadata_presenter/next_page.rb
```diff
@@ -11,17 +11,53 @@ def find
...
     end
 
     def return_to_check_your_answer?
-      session[:return_to_check_your_answer].present?
+      session[:return_to_check_your_answer].present? &&
+        no_branch_ahead_for_current_page_and_user_did_not_changed_answers?
+    end
+
+    def no_branch_ahead_for_current_page_and_user_did_not_changed_answers?
+      no_branch_ahead_for_current_page? && user_did_not_changed_answers?
```
brenetic, line 30:
```suggestion
      components_not_used_elsewhere? && answers_unchanged?
```

yeah difficult naming. it's just checking the current pages components aren't used in another branch later in the flow right?

tomas-stefano: Yes, I wasn't happy with the naming hahaha. I like the suggestion for `answers_unchanged` but the components not used anywhere maybe don't reflect entirely what is happening. Maybe could be `components_not_used_for_branching?`. Wdyt?

app/models/metadata_presenter/next_page.rb
tomas-stefano: 💯 Although this will require change when we decide to do the magic return to cya in the future far away.

app/models/metadata_presenter/next_page.rb
brenetic: 👍  yup, let's go with that :)

brenetic (approved):
🚀

app/models/metadata_presenter/next_page.rb
```diff
@@ -11,17 +11,53 @@ def find
...
       service.pages.find { |page| page.type == 'page.checkanswers' }
     end
 
     def return_to_check_your_answer?
-      session[:return_to_check_your_answer].present?
+      session[:return_to_check_your_answer].present? &&
+        components_not_used_for_branching_and_user_did_not_changed_answers?
+    end
+
+    def components_not_used_for_branching_and_user_did_not_changed_answers?
```
brenetic, line 29:
```suggestion
    def components_not_used_for_branching_and_answers_unchanged?
```

tomas-stefano: Done

brenetic (approved):
🚀

PR #179, opened 2021-09-27
Use grid and route models

brenetic:
This adds a couple of models which are used to:

- Traverse all potential routes for a given service
- Build a grid based object which serves the service flow page

## Route

The Route model is used to traverse a potential path through a service's
logical flow. It will follow the default next destination for a given
page.

The first Route always traverses from the start page.

When it gets to a branch object it will take the first conditional
destination and follow that through until the end of the form, or until
an exit page is reached along that route.

It will then create additional Route instances which will traverse from
the conditional destination.

Route objects have no knowledge of any other Route objects. They simply
represent a possible path that can be taken through a service.

Each Route object knows the row and column location of where it starts
to traverse from. Since the Route has no knowledge of what is in other
Routes, the Grid object sometimes has to update the starting row for a
given Route.

## Grid

**Important note: All rows and columns references here start at index 0.**

The Grid model is responsible for taking all the potential Routes
through a service and building out a matrix/grid (array of arrays) that
represents the rows and columns layout of the service flow.

It first builds a matrix based upon the total number of potential routes
by the maximum number of potential columns. The latter is figured out by
taking the Route with the highest starting column number and adding on
the total number of unique flow UUIDs that that Route traversed.

Always starting with the Route that is traversing from the start page,
the Grid first iterates over each UUID and updates the column property
of the object. It will always update the column number with the highest
should it come across objects that it has seen before.

Once columns have been added for each flow object, the UUIDs are
iterated over again and if it is the first time that object was seen then
the row property for that object is updated. Otherwise it is skipped.

After the above two steps any flow objects in that still have nil value
for their row and column properties are considered detached from the
main flow.

Once the rows and column properties have been updated, each object is
then placed in the correct position in the `@ordered` matrix grid that was
created earlier.

Route rows need to also be updated when iterating over each branch object.
All the potential destinations that are exiting from a given branch are
looked at and any that already exist in any of the rows above are
ignored. Then the Route objects that have their starting point from any
of the remaining destinations are found and their rows updated in
relation to the current row that the branch object is on. Doing this
prevents any unwanted Spacer objects being left in columns directly
after a branching point.

There is one caveat to the above paragraph and that is when additional
space is required for the conditional questions and answers that are
displayed directly after each branching point.

The UI treats each 'or' expression type for any given conditional as
something that requires it's own line and arrow. Therefore an additional
Spacer object needs to be injected into the corresponding column
directly after a branch if any of it's conditionals is an 'or'. Any
conditionals which are 'and' are still grouped together and therefore do
not need to be separated out in the same way.

Lastly any Spacer objects that are after the last position of a Flow
object for any given column are removed.

The Grid object also will return a flattened ordered flow array
containing branches and pages and a flattened ordered pages array
containing only pages. These are used for building the Destination lists
in the Change Destination action as well as the Questions list on the
Edit Branch page. Neither of these require Spacer objects, they just
need to be in the correct relative order of the service flow.

## Exceeding total number of flow objects or potential routes

We have added a couple of ActiveSupport::Notifications for the Route and
Grid objects. These are here to warn about times when iterating loops
may have continued to be infinite. We calculate the total number of
potential routes as well as the total number of objects that need to be
iterated over and break if those numbers are ever exceeded. We will
monitor this to see if it is appropriate.

## Example

The diagrams below are representational only. The actual objects are placed
in the grid only _after_ the row and column numbers have been calculated.

Screenshot 2021-09-25 at 23 22 14

Given the above flow there are four potential Routes that will be generated for it:

Screenshot 2021-09-25 at 23 22 20

Each Route object has a starting row, 0, 1, 2, 3 respectively and also a corresponding starting column, 0, 2, 5, 5. These are naive points of reference but are used by the Grid object as a starting point only.

The grid object generates the initial grid matrix:

Screenshot 2021-09-27 at 18 28 48

Then iterates of the Route that starts from the start page and places each object:

Screenshot 2021-09-27 at 18 32 27

Since this is the first row it will place objects in whatever column it comes to first.

The next row will the iterate in the same way. If it gets to an object that has been previously iterated it will update the column number for that object accordingly:

Screenshot 2021-09-27 at 18 34 44

P3 and P4 will have their column properties updated to represent their later
positions. Because Branching Point 2 is part of this particular Route, the Grid model will check the exit destinations for the branch and compare the starting row numbers for each Route object that starts from those destinations. If any are matched it will update the row numbers for each Route relative to the row Branching Point 2 is on.

In this case Route 2 would have been created with the starting row `2`, but since P3 exists in a row above and was also one of the potential exit destinations for Branching Point 2, it's row will be update to be `1`.

Route 2 is then iterated over:

Screenshot 2021-09-27 at 18 42 00

The only new object is P7. However P3 and P4 exist as potential paths and are in later columns so their column positions are updated accordingly.

Lastly Route 3 is iterated over:

Screenshot 2021-09-27 at 18 47 10

P8 and P9 are placed in the correct column. P4 is not updated as the column position for it has not increased.

Once the row and column coordinates are properly setup the Grid then adds each object to the correct relative points in the matrix. Then any Spacers that are not required are removed from the matrix:

Screenshot 2021-09-27 at 18 49 39

Lastly the Grid will go back over the columns directly after any Branching Point and add additional Spacer objects should the branch have any conditionals of type 'or'.

Simples.


tomas-stefano (approved):
🚀

PR #187, opened 2021-10-20
Conditional uuids

brenetic:
The conditional object in a branch needs to have a UUID so that we can
identify which conditional to update when adding a page after a branch.

Publish 2.8.0


njseeto (approved):
🚀

PR #188, opened 2021-10-21
Update all fixtures to include the conditional uuids

tomas-stefano:
This is the only way we can do BDD on the editor for adding a page
after a branch


brenetic (approved):
🚀

PR #189, opened 2021-12-02
Allow traversing of objects later in the flow 

brenetic:
Previously we were not considering objects that would have appeared
later in the flow if they had been previously traversed. i.e if an object
is pointed to twice we would only be considering the first time it
appeared in the flow.

Update this to only traverse if a route has not been traversed as
opposed to each individual object.

Add Branching Fixture 9 with examples of the above

branching_9

Publish 2.10.0


njseeto (approved):
🚀

PR #190, opened 2021-12-08
Make space for visually stacked branches 

brenetic:
We do not allow branches to be stacked in terms of data integrity but we
do allow them to be visually stacked. In these instances we need to make
space for branch expressions ('and' and 'or' types) so that the frontend
has the ability to display the question and answers in the pages flow.

At the same time this change also takes into account secondary routes
that may or may not link to pages and branches that have been previously
linked to earlier in the flow. In those instances we want to take the
later column, and under some circumstances we want to also update the
row numbers.

In order to accomplish this we have broken out the grid object into smaller
more easily testable models which wrap Coordinates, Rows and Columns.

Also add another branching fixture:

branching_10

Publish 2.11.0


app/models/metadata_presenter/coordinates.rb
```diff
@@ -0,0 +1,59 @@
...
+          return uuid
+        end
+      end
+    end
+
+    def position(uuid)
+      positions[uuid]
+    end
+
+    def occupied?(column, row, new_object_uuid)
```
tomas-stefano, line 38: Cool 👍

app/models/metadata_presenter/coordinates.rb
```diff
@@ -0,0 +1,59 @@
...
+    private
+
+    attr_reader :flow
+    attr_writer :positions
+
+    def setup_positions
+      flow.keys.index_with { |_uuid| { row: nil, column: nil } }
+    end
+  end
+end
```
tomas-stefano, line 59: Great code 🚀

tomas-stefano (approved):
🚀 🚢

njseeto (approved):
🏅 🚀 💯

PR #191, opened 2021-12-13
Service name validation revisited

tomas-stefano:
[Trello card](https://trello.com/c/gEognZYS/2133-allow-special-characters-like-apostrophe-in-the-form-name)

## Context

We need to accept service names with single quotes/apostrophe so adding specs to make sure the slugs are generated as expected

### DIFF
--- a/spec/models/service_spec.rb
+++ b/spec/models/service_spec.rb
@@ -7,7 +7,9 @@ RSpec.describe MetadataPresenter::Service do
         'grogu' => 'grogu',
         'Darth Vader' => 'darth-vader',
         'emperor-palpatine' => 'emperor-palpatine',
-        'princess Leia' => 'princess-leia'
+        'princess Leia' => 'princess-leia',
+        "Darth Vader Children's" => 'darth-vader-childrens',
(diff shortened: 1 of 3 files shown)


spec/models/service_spec.rb
```diff
@@ -7,7 +7,9 @@
         'grogu' => 'grogu',
         'Darth Vader' => 'darth-vader',
         'emperor-palpatine' => 'emperor-palpatine',
-        'princess Leia' => 'princess-leia'
+        'princess Leia' => 'princess-leia',
+        "Darth Vader Children's" => 'darth-vader-children-s',
+        'Darth Vader Children’s' => 'darth-vader-children-s'
```
njseeto, line 12:
```suggestion
        "Darth Vader Children's" => 'darth-vader-childrens',
        'Darth Vader Children’s' => 'darth-vader-childrens'
```
should the urls look like this?
So it is similar to the `claim-for-costs-of-a-childs-funeral.form.service.justice.gov.uk` url.

njseeto (approved):
🚀

PR #193, opened 2021-12-14
Stop grid object overwrite

brenetic:
Previously the Grid did not check whether an object with a different
uuid existed in the location that is is attempting to place another Flow
object.

Check whether the space is already occupied and if so insert a Spacer
before placing the new object.

This is not currently able to cater for the edge cases which involve the
ordering of destinations exiting from a branch. However it is more
important at this point that _all_ required objects that are linked to
appear on the page instead of becoming detached.

Also cater for branches being more than one row above objects.
In some instances it is possible for a Flow object to be in the same
column as a branch object which is above it. However it might be more
than one row above it.

Make sure there are enough spaces in between for the expressions. At the
same time do not add more spaces than necessary


njseeto (approved):
🚀

tomas-stefano (dismissed):

PR #194, opened 2021-12-15
Add warning object

njseeto:
[Trello](https://trello.com/c/657G0Env/2121-add-warning-for-deleting-cya-or-confirmation-page-4-7)

### Add warning object
The Warning object will be used on the Pages Flow to warn the user if they do not have a CYA or a Confirmation page in the main flow.
The Warning object should always be on the top row, in the last column of the grid.
The Warning object should not be present in the unconnected pages section.
 
### Version 2.13.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ All notable changes to this project will be documented in this file.
 ### Changed
 ### Fixed
 
+## [2.13.0] - 2021-12-15
+
+### Changed
+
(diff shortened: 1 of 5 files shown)


tomas-stefano (commented):
Looks great, just minor comments 💯

brenetic:
great stuff, nat!

agree with @tomas-stefano about the naming of the methods. it is specifically about the CYA and Confirmation pages and not about generic pages

brenetic (approved):
🚀

PR #196, opened 2021-12-20
Fix flow objects pointing backwards

brenetic:
## Do not force the use of the latest column or row 

Previously we had been pushing the objects right as we add them to grid.
This does in fact make for a slightly better viewing experience in some
instances.

However when there are times when an editor has chosen to
link back to a previous object in the flow, the grid was attempting to
shift everything forward to the very latest column. This resulted in
broken flows.

Instead just use the previously set column number and allow the frontend
to draw the line to that object. The only exceptions to this rule are
the CYA and Confirmation pages.

In order to enable the correct positioning of objects that live in the
same column as branches we need to cater for the total number of
potential conditions that that branch is required to draw lines for.
This includes the additional spacers required for future 'OR'
conditionals.

We now generate a separate spacer object that holds the column and row
numbers for each branch conditional. This enables us to check how many
rows the branch _actually_ needs in order to have the required visual
space before the next object can be placed.

## Additionally

- Check that UUIDs that are not meant to be in the main flow do not appear 
   when calling `flow_uuids`.
- Add branching fixture 11 which includes backward linking as well as detached
   objects
- Update branching fixture 9 and 10 names
- 2.14.0

branching_11


njseeto (approved):
🚀

PR #197, opened 2021-12-20
Update flow object row based on branch spacers

brenetic:
When a branch object row is calculated we also calculate the positions
of the required Spacer objects that have to go below it in the grid.

Now we associate those Spacers to a specific object UUID. Doing this
enables us to check whether the objects in the column following a branch
need to be moved to a higher row.

The Spacers for each branch are used to hold the text associated with
any conditional logic question and answer which dictates the visual
flow. If we did not update the row numbers then this can lead to objects
being left on rows further up the page.

Publish 2.14.1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ All notable changes to this project will be documented in this file.
 ### Changed
 ### Fixed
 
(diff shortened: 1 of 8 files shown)


njseeto (approved):
🚀

PR #200, opened 2022-01-06
Check detached submitting pages

brenetic:
Whenever a change to the metadata of a service affects the checkanswers
or confirmation pages we need to check whether they have been detached
from the main flow. If so then we are most likely going to show a
warning message to the user that they will be unable to submit their
form.

We also have to surface the Service object inside the Grid model because
the Editor will need to compare grids for two different types of services; pre
and post user changes.

Publish 2.15.1


njseeto (approved):
🚀

PR #204, opened 2022-01-13
Update default heading for email component

njseeto:
Following testing the team have decided to change the default heading from
Question to Email address question.
This is to avoid confusion in the event a user adds a text and email component
on a multiple question page.

Also, re-arrange the order of components on a multiple question page so the
text based components are grouped.

Screenshot 2022-01-13 at 09 42 49


brenetic (approved):
🚀

PR #206, opened 2022-01-14
Update version fixture

njseeto:
Replace existing email (text) component with an actual email component.
The old email component had custom validations for error messages, these have been moved to the `parent-name` component and specs have been updated to reflect this.

### 2.15.6


brenetic (approved):
🚀

PR #209, opened 2022-02-08
Remove excess Pointer objects in detached grids

brenetic:
The rules for building grids which are used in the detached section of
the Editor is a little different from the one used for the main flow.

The main flow usually takes the column number of the object that was
previously positioned, unless it's a checkanswers page or confirmation
page, in which case it will use the latest column number.

For detached grids we want to trim Pointer objects back so each row ends
in the first Pointer it comes across. However in many instances the
object that is being pointed to might be on a completely different row
and either later or earlier in the flow. This is mostly a problem when
it comes to branching point destinations.

In those instances we check whether there should be a Pointer for that
object. If so then we replace the Spacer with a Pointer.

Conversely if there are any rows that only contain Spacers and Pointers,
and none of the Pointers are a branch destination from the column
directly before, then we do not really want that row at all. In these
cases swap any Pointers for Spacers and allow the trim_spacers method to
remove the rows at the end of building the grid.

Publish 2.15.9

# Before

Screenshot 2022-02-07 at 13 03 20

# After

Screenshot 2022-02-08 at 14 28 57

### DIFF
--- a/app/models/metadata_presenter/grid.rb
+++ b/app/models/metadata_presenter/grid.rb
@@ -41,7 +41,7 @@ module MetadataPresenter
       set_row_numbers
       add_by_coordinates
       insert_expression_spacers
-      trim_pointers unless main_flow.empty?
+      trim_pointers unless main_flow.empty? # only used by detached grids
       trim_spacers
       insert_warning if main_flow.empty?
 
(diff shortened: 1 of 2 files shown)


chrispymm (approved):
All looks good to me.

njseeto (approved):
🚀

PR #210, opened 2022-02-09
CYA and Confirmation page column placement

njseeto:
[Trello](https://trello.com/c/K0G1FddV/2264-confirmation-page-sometimes-appears-before-check-your-answers-page-48-59-s-m)
### CYA and Confirmation page column placement
During grid placement, check whether a check answers page exists.
If it exists, ensure the confirmation page is in the next column.

## Before
153198480-ca23b810-6143-47e0-858e-1f8dea4cb2e1

## After
Screenshot 2022-02-09 at 12 10 24

### DIFF
--- a/spec/models/column_number_spec.rb
+++ b/spec/models/column_number_spec.rb
@@ -72,15 +72,113 @@ RSpec.describe MetadataPresenter::ColumnNumber do
         end
       end
 
-      context 'when checkanswers or confirmation page' do
-        let(:new_column) { 10 }
+      context 'when checkanswers page' do
+        # branching fixture 10 CYA page
+        let(:uuid) { 'da2576f9-7ddd-4316-b24b-103708139214' }
+        let(:positions) do
+          {
+            uuid => {
+              column: 8,
+              row: 1
+            }
+          }
+        end
 
-        # branching fixture 10 CYA and Confirmation pages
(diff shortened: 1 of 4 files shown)


spec/models/column_number_spec.rb
```diff
@@ -72,15 +72,62 @@
...
+            }
+          end
+          let(:latest_column) { 10 }
+
+          it 'returns the latest column' do
+            expect(column_number.number).to eq(latest_column)
+          end
+        end
+
+        context 'when confirmation page is not present' do
```
brenetic, line 97: there are at least two contexts for this i think. `latest_column` will return the higher of the `existing_column` or the `new_column`. in your example on line 100 it will always return the `new_column` as an existing checkanswers column does not exist.

so there probably needs to be contexts when there is no existing checkanswers, where there is an existing checkanswers but the number is less than the `new_column` and then where there is an existing checkanswers but the number is greater than the `new_column`.

spec/models/column_number_spec.rb
```diff
@@ -72,15 +72,113 @@
...
+
+        context 'when existing column is less than new column' do
+          let(:new_column) { 10 }
+
+          it 'returns the new column' do
+            expect(column_number.number).to eq(10)
+          end
+        end
+
+        context 'when existing column is more than new column' do
```
brenetic, line 95:
```suggestion
        context 'when new column is more than existing column' do
```
```suggestion
        context 'when existing column is more than new column' do
```

brenetic (approved):
🚀

PR #211, opened 2022-02-14
Fix branching point spacers

njseeto:
[Trello](https://trello.com/c/qCUjCw7r/2256-bug-page-objects-overlap-branching-lines-s-m)
## Fix branch spacers to allow for more branches pointing to the same page
At present, the number of branch spacers is calculated by the number of exiting destinations on the branching point and uses the destination UUID as an identifier. This means it is possible the number of spacers below a branching point could be incorrect if more than one conditional points to the same page, leading to page objects sitting on top of branch lines.

### Before
branching 12 before

### After
branching 12 after

## Allow branch spacers to account for 'OR' conditionals
We should only count unique uuids if we have 'OR' conditions in a branch so we do not add too many branch spacers in the following column.

### Before
branching without OR

### After
branching after
 
## Bump 2.15.10


spec/models/coordinates_spec.rb
```diff
@@ -4,6 +4,28 @@
   let(:service) { MetadataPresenter::Service.new(latest_metadata) }
   let(:uuid) { 'ad011e6b-5926-42f8-8b7c-668558850c52' } # Page N
 
+  describe 'initialising the branch spacers' do
+    context 'when branching point does not have OR conditions' do
+      let(:latest_metadata) { metadata_fixture(:branching_12) }
```
brenetic, line 9: if this is the only place branching 12 is used, and you only need it because it does not have an OR condition, then any of the branching fixtures except `branching` would do

njseeto: Happy to remove branching fixture 12 as it is only used here, although none of the other fixtures have revealed this bug in the past and I would like to specifically test this bug. I can make it work with some manipulation of fixture 10 but it would mean a bit of test setup, I can give it a go.

njseeto: Turned out fixture 10 was fine to use without adjustment as it had a similar set up to fixture 12 🎉 
Have now done this.

brenetic (approved):
Great stuff, nat!

🚀

PR #212, opened 2022-02-18
Add `contains` and `does_not_contain` operators

njseeto:
[Trello](https://trello.com/c/5LKINamZ/2334-condition-using-a-checkbox-operator-is-incorrect-is-isnt-instead-of-contains-doesnt-contain-38-s)

### Make the contains operator work
Calling `classify` on the  `contains` operator in the `operator.rb` `klass` method, removes the 's'. Adding a rails inflector transforms the `contains` into `ContainsOperator` class to evaluate the condition.
 
### Bump presenter 2.15.11

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ All notable changes to this project will be documented in this file.
 ### Changed
 ### Fixed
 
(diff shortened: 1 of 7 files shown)


brenetic (approved):
🚀  awesome

PR #213, opened 2022-03-17
Add ability to check if a page is a multiple question page

njseeto:
Relates to [Trello](https://trello.com/c/JY7GLf37/2343-in-branching-screen-condition-question-does-not-use-optgroup-to-group-q-from-multi-q-pages-67-m)

- Add ability to check if a page is a multiple question page
- Bump presenter 2.15.12


brenetic (approved):
🚀

PR #215, opened 2022-04-05
Expose previous destination uuids

brenetic:
This adds previous flow uuids on the route model for any objects that it
traverses. It will also add a conditional uuid should that be
applicable.

For any given route object, any branches that are part of that
particular route, the first conditional will be traversed as normal and
the previous flow uuid and conditional uuid set for that object.

Each additional branch destination and default next for that branch get
their own route object which will traverse that path and set previous
flow uuids and conditional uuids accordingly.

The route model exposes the previous flow uuid and conditional uuid for
each flow object that is traversed for any given route.

These now need to be set for any given grid, main flow or detached.
Similar rules apply for when flow objects are placed on the grid; the
first flow object takes precedence over any preceding objects. This is
to say that only the first object that is placed on the grid will have
the ability to be moved as the previous flow uuid and optional
conditional uuid will not be overwritten by any following objects.

Also:

### Additional Route object tests for column and row incrementation

The first destination of a branch is traversed by the parent route
object. Each subsequent destination and the branch default next is done
by separate route objects.

Each of these route objects is initialised with an incrementing column
and row number.

### Additional test for setup of positions object in Coordinates model

The Coordinates model initialises with a positions object which caters
for row, column, previous flow uuid and potential conditional uuid if
the object in question is a branch destination.

### Publish 2.16.0

### DIFF
--- a/app/models/metadata_presenter/coordinates.rb
+++ b/app/models/metadata_presenter/coordinates.rb
@@ -61,7 +61,9 @@ module MetadataPresenter
     attr_writer :positions, :branch_spacers
 
     def setup_positions
-      service.flow.keys.index_with { |_uuid| { row: nil, column: nil } }
+      service.flow.keys.index_with do |_uuid|
+        { row: nil, column: nil, previous_flow_uuid: nil, conditional_uuid: nil }
+      end
     end
 
     # This also takes into account the 'OR' expressions which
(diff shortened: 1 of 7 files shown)


mattempty (approved):
🚀

chrispymm (approved):
🚀

PR #216, opened 2022-04-08
Add responsive meta viewport tag and version bump

chrispymm:
This adds the correct `viewport` meta tag from the design system so that the interface displays correctly on devices.

image

Before:
image

After:
image


brenetic (approved):
🚀

PR #217, opened 2022-04-21
Set confirmation page column coordinate to after the CYA page

njseeto:
[Trello](https://trello.com/c/c4I4IvM3/2410-sometime-the-flow-view-shows-confirmation-before-cya-and-other-bugs)
- If confirmation and CYA pages exist, deliberately set the confirmation page to be after the CYA page.
- Bump to 2.16.2

### Before
before

### After
after

### DIFF
--- a/app/models/metadata_presenter/grid.rb
+++ b/app/models/metadata_presenter/grid.rb
@@ -214,7 +214,13 @@ module MetadataPresenter
         position = coordinates.positions[uuid]
         next if detached?(position)
 
-        column = position[:column]
+        # Update confirmation page column co-ordinate
+        column = if confirmation_and_cya_pages_present? && service.confirmation_page.uuid == uuid
+                   coordinates.positions[service.checkanswers_page.uuid][:column] + 1
+                 else
+                   position[:column]
+                 end
+
         row = position[:row]
         insert_spacer(column, row) if occupied?(column, row, uuid)
         @ordered[column][row] = get_flow_object(uuid)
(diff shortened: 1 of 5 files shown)


app/models/metadata_presenter/grid.rb
```diff
@@ -225,6 +231,12 @@ def detached?(position)
       position[:row].nil? || position[:column].nil?
     end
 
+    def confirmation_and_cya_pages_present?
+      service.confirmation_page.present? &&
+        service.checkanswers_page.present? &&
+        !coordinates.positions[service.checkanswers_page.uuid][:column].nil?
```
brenetic, line 237:
```suggestion
        coordinates.positions[service.checkanswers_page.uuid][:column].present?
```

app/models/metadata_presenter/grid.rb
```diff
@@ -215,6 +215,12 @@ def add_by_coordinates
         next if detached?(position)
 
         column = position[:column]
+
+        # Add confirmation page column co-ordinate
```
brenetic, line 219:
```suggestion
        # Update confirmation page column co-ordinate
```

brenetic: think maybe an if else might be good

app/models/metadata_presenter/grid.rb
brenetic: if only the confirmation page is present i assume the grid just places the confirmation page wherever the column model calculated it?

brenetic:
i think we can have some tests around this, maybe in the grid _spec

app/models/metadata_presenter/grid.rb
njseeto: yep correct, the grid will place the confirmation page where the column model has calculated.

brenetic (approved):
🚀

PR #220, opened 2022-05-05
Minimum validator default metadata

brenetic:
This adds the default metadata for the minimum validation that the Editor adds to the metadata. Defaults to "0".

Also some of the default metadata did not have keys associated with them as they were missing either '_id' or '_type'. Fall back to the filename if that is the case.

Also if any schema has definitions internally then iterate over those and also add them as inidividually retrievable schemas when doing validation.


njseeto (approved):
🚀

PR #221, opened 2022-05-05
Lock json-schema version to 2.8.1

brenetic:
There are changes in json-schema which incorporate the draft JSON Schema
version 3 which causes validation failures in our current schemas.

These need to be investigated to ascertain the differences and whether
we need to adjust anything. In the meantime lock the json-schema gem at
2.8.1


njseeto (approved):
🚀

PR #222, opened 2022-05-05
Publish 2.16.3

brenetic:
This release includes:

- Added Minimum validation for the number component
- Added default metadata for minimum validation
- Use file path as fallback for naming default metadata keys
- Add schema sub definitions as their own schemas to the JSON validator
- Lock json-schema to version 2.8.1


njseeto (approved):
🚀

PR #226, opened 2022-05-09
2454 confirmation template width bug

stevenburnell-moj:
- Fix small design bug (width Confirmation)
- Silenced JS error message (was pulling unwanted Editor code)

**Visual was:**
Screenshot 2022-05-09 at 08 45 12

**Visual now:**
Screenshot 2022-05-09 at 08 45 23

### DIFF
--- a/app/views/layouts/metadata_presenter/application.html.erb
+++ b/app/views/layouts/metadata_presenter/application.html.erb
@@ -30,7 +30,6 @@
     </div>
     <%= render template: 'metadata_presenter/footer/footer' %>
 
-    <%= javascript_pack_tag 'application' %>
     <%= javascript_pack_tag 'govuk' %>
   </body>
 </html>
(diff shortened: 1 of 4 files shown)


app/views/layouts/metadata_presenter/application.html.erb
```diff
@@ -30,7 +30,6 @@
     </div>
     <%= render template: 'metadata_presenter/footer/footer' %>
 
-    <%= javascript_pack_tag 'application' %>
     <%= javascript_pack_tag 'govuk' %>
```
chrispymm, line 33: Are we certain the application js is not needed?  I understand that it *shoudln't* be needed, but just want to make sure we don't break forms in production...

stevenburnell-moj: Yes, I believe so.

Application.js basically did this:
- Pull in jQuery stuff
- Pull in Showdown (markdown editing support)
- Pull in Editor index.js for Editor JS support
- Pull in Editor CSS

None of that we should need since it is all Editor specific and, the only things we want in Runner are the GDS CSS

chrispymm (approved):

PR #229, opened 2022-05-11
Min and Max Length default metadata

brenetic:
This adds the min and max length default metadata for string component
validations.


njseeto (approved):
🚀

PR #230, opened 2022-05-16
Change validation number types to string

brenetic:
While we are indeed asking a user to enter a number for certain
validations we always store the value as a string.

Validation of that number happens at application level as we coerce the
string value to a Float.


njseeto (approved):
🚀

PR #237, opened 2022-05-20
Check number value for necessary validations

brenetic:
Similar to the date component, the default metadata for a number
component already includes the 'Number' validation. Therefore do not
bother validating the minimum or maximum unless the number validation is
valid.


njseeto (approved):
🚀

PR #238, opened 2022-05-27
Autocomplete schemas

njseeto:
[Trello](https://trello.com/c/PLyp8ebs/2588-autocomplete-metadata-presenter-schemas-and-templates)

- Add schemas and metadata for autocomplete component
- Bump presenter to 2.16.11

### DIFF
--- a/default_metadata/component/autocomplete.json
+++ b/default_metadata/component/autocomplete.json
@@ -1,0 +1,12 @@
+{
+  "_id": "component.autocomplete",
+  "_type": "autocomplete",
+  "errors": {},
+  "hint": "",
+  "options": [],
+  "name": "component-name",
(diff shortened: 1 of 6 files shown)


default_metadata/component/autocomplete.json
```diff
@@ -0,0 +1,25 @@
...
+      "label": "Option",
+      "value": "value-2"
+    }
+  ],
+  "name": "component-name",
+  "legend": "Question",
+  "validation": {
+    "required": true
+  }
+}
```
brenetic, line 25: i don't think we need options to be prefilled here. these are dynamic that are only loaded by the frontend, otherwise it's just a select list

brenetic (approved):
🚀

PR #241, opened 2022-06-07
Google analytics tags

brenetic:
This adds the script tags for analytics to the head of the application
layout. We initially support Google Analytics at this time. There are
three types:

- UA (Universal Analytics)
- GTM (Google Tag Manager)
- GA4 (Google Analytics 4)

Each one injects a different script into the head. In the case of Google
Tag Manager it also adds a no-script tag to the beginning of the body.

### DIFF
--- a/app/views/metadata_presenter/analytics/_google.html.erb
+++ b/app/views/metadata_presenter/analytics/_google.html.erb
@@ -1,0 +1,5 @@
+<% analytics.each do |analytic| %>
+  <% if ENV[analytic].present? %>
+    <%= render partial: "metadata_presenter/analytics/#{analytic.downcase}", locals: { analytic_tag: ENV[analytic] } %>
+  <% end %>
+<% end %>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ All notable changes to this project will be documented in this file.
 ### Changed
 ### Fixed
 
+## [Unreleased]
+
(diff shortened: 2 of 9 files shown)


app/views/metadata_presenter/analytics/_google.html.erb
```diff
@@ -0,0 +1,12 @@
...
+    <script async src="<%= "https://www.googletagmanager.com/gtag/js?id=#{ENV[analytic]}" %>"></script>
+    <script>
+      window.dataLayer = window.dataLayer || [];
+      function gtag(){dataLayer.push(arguments);}
+      gtag('js', new Date());
+
+      gtag('config', <%= ENV[analytic] %>);
+    </script>
+  <% end %>
+<% end %>
```
stevenburnell-moj, line 12: Pretty sure there are different snippets required for different Analytics services (e.g. depending on the account type in use).

I copied some examples down in this ticket https://trello.com/c/nJgqBCn1/2565-google-analytics-kick-off
(see comments)

Is this just a starting point or meant to be something that covers the different types of Analytics available from Google?

brenetic: i thought there were entirely different snippets as well, but having now gone to the analytics dashboard to generate the snippet for the three different types they all use the same script tags 🤷‍♀️

stevenburnell-moj (approved):

PR #242, opened 2022-06-07
Allow users to disable non essential cookies 

brenetic:
This adds the cookie banner to the top of forms allowing the user to
either accept or reject non essential cookies.

It will set an additional essential cookie on the clients device which
takes the form `analytics-<parameterized-form-name>` and has a value of
either `accepted` or `rejected`.

Currently we do not allow the user to change their preferences. This
will be done in a future story.


stevenburnell-moj (approved):

njseeto (approved):
LGTM 🚀

PR #243, opened 2022-06-09
Correctly traverse pages based on user answers

brenetic:
The object responsible for traversing user answers can take a page
which tells it when to stop. If none is provided it would previously
take the last page in the `pages` array of the metadata. This was
perfectly fine before branching as the `pages` array was the thing
used to dictate the order of a form which always ended in a
Confirmation page.

However now branching uses a `flow` object system to dictate what a
user should see next based on their answers, the `pages` array order
no longer matters. The last page is simply the last page created by
the user.

Therefore we should traverse the user answers all the way to end of a
journey based upon their answers. In the instances of a page being
supplied we traverse up to and not including that page. If none is
provided then we go all the way until there is no longer the ability to
find another page. These instances are:

- end of the flow (i.e a confirmation page)
- an exit page
- a page with no default next which usually only happens if the form is
  currently not been finished its creation by the Editor user
  
Additionally get the start page by type instead of picking the first page
in the `pages` array.

Publish 2.17.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ All notable changes to this project will be documented in this file.
 ### Changed
 ### Fixed
 
+## [2.17.0] - 2022-06-09
+
+### Fixed
+
+- Fixed issue where pages traversed based on user answers were incomplete and therefore
+  missing from submission payloads and generated PDF files
+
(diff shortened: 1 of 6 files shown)


mattempty (dismissed):
🚀

PR #244, opened 2022-06-09
Fix analytics id strings

brenetic:
When passed into the JS functions the GA tags need to be string literals.

Publish 2.16.14


mattempty (approved):
🚀

PR #249, opened 2022-06-27
Add MoJ Forms global analytics

brenetic:
This injects the global analytics ID for the GA4 used by the MoJ Forms
team.

Since MoJ Forms will be using GA4 we do not want to overwrite a form
owners GA4 or vice versa. We only need to call the gtag function with
the additional measurement ID, as opposed to load the entire JS script
again.

Also changed `analytics_tag` to `measurement_id` as that is what Google
refers to it as.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ All notable changes to this project will be documented in this file.
  - Update button label for check answers page to 'Submit'
  - Move button labels into locales file
 
+### Added
+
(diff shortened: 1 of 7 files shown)


njseeto (changes requested):
Code looks good, just one small comment re: changelog

njseeto (dismissed):
🚀

njseeto (approved):
🚀

PR #250, opened 2022-06-27
Update check answers page button text and move button labels into locales 

chrispymm:
Changes the button text on the check answers page to 'Submit'

This label is now within the locales file.

Also moved the 'Start Now' and 'Continue' button labels into the locales file.

image


brenetic (approved):
🚀

PR #256, opened 2022-07-12
Move editor specific functionality of autocomplete into Editor

njseeto:
[Trello](https://trello.com/c/mU4YFw6r/2688-autocomplete-check-for-previously-uploaded-items-on-the-component-view)

The upload options button on the autocomplete component, will only be seen in the Editor. The runner does not need to know about this functionality. As such, we have moved this part of the autocomplete template out of the presenter and into a partial in the Editor app.

### Bump to version 2.17.8
This version also includes the removal of Graphviz dependency and associated rake task found in this [PR](https://github.com/ministryofjustice/fb-metadata-presenter/pull/255)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,12 @@ All notable changes to this project will be documented in this file.
 ### Changed
 ### Fixed
 
-## [Unreleased]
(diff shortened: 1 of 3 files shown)


brenetic (approved):
🚀

PR #257, opened 2022-07-18
2727 global analytics

stevenburnell-moj:
https://trello.com/c/h3595yEp/2727-show-cookie-banner-when-global-analytics-is-configured-for-a-form

Cookie Banner controlling update and global (MoJ Forms) analytics always present (if user permission given).

### DIFF
--- a/app/views/layouts/metadata_presenter/application.html.erb
+++ b/app/views/layouts/metadata_presenter/application.html.erb
@@ -38,8 +38,7 @@
       </main>
     </div>
     <%= render template: 'metadata_presenter/footer/footer' %>
-
-    <%= javascript_pack_tag 'application' %>
+    <%= javascript_pack_tag 'runner_application' %>
     <%= javascript_pack_tag 'govuk' %>
   </body>
 </html>
--- a/app/views/metadata_presenter/analytics/_ga4.html.erb
+++ b/app/views/metadata_presenter/analytics/_ga4.html.erb
@@ -1,16 +1,8 @@
-<!-- Google Analytics 4 -->
+<!-- Form Owner Google Analytics 4 -->
 <% if Rails.application.config.respond_to?(:global_ga4) %>
-<script>
-  gtag('config', '<%= measurement_id %>');
-</script>
-<% else %>
-<script async src="https://www.googletagmanager.com/gtag/js?id=<%= measurement_id %>"></script>
-<script>
-  window.dataLayer = window.dataLayer || [];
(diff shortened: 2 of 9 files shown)


app/views/metadata_presenter/analytics/_ga4.html.erb
```diff
@@ -1,16 +1,8 @@
...
-  window.dataLayer = window.dataLayer || [];
-  function gtag(){dataLayer.push(arguments);}
-  gtag('js', new Date());
-
-  gtag('config', '<%= measurement_id %>');
-</script>
+  <%# Global Google Analytics (GA4) should be included before this point so we only need a config setting %>
+  <script>
+    gtag('config', '<%= measurement_id %>');
+  </script>
```
brenetic, line 6: this would mean that the global GA4 _has_ to exist? in other words, it would not be possible for us to remove the global analytics but to allow a form owner to use GA4.

app/views/layouts/metadata_presenter/application.html.erb
```diff
@@ -38,8 +38,7 @@
       </main>
     </div>
     <%= render template: 'metadata_presenter/footer/footer' %>
-
-    <%= javascript_pack_tag 'application' %>
+    <%= javascript_pack_tag 'runner_application' %>
```
brenetic, line 41: the preview mechanism in the editor offloads the layout responsibility to the presenter. would this be ok in those instances?

stevenburnell-moj: Yes. It's a little disappointing I cannot easily show the partner action in this PR but there is are additional changes (see other PRs in waiting) to both Editor and Runner that make sure the requested file is in place. It's not the nicest solution but it will prevent the JS error we get in Preview that blocks any valid JS from running after this point.

app/views/metadata_presenter/analytics/analytics.html.erb
```diff
@@ -1,7 +1,11 @@
+<%# MoJ Forms Google Analytic Reporting %>
```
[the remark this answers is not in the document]
stevenburnell-moj: Developer (template level) comment intentional because I'm already doing the HTML level one inside the app/views/metadata_presenter/analytics/_global_analytics.html.erb template. This one, like the `<%# Analytic reporting set by form owners %>` that also features in this partial, is an attempt to avoid developer confusion over what the partial is doing (and calling).

brenetic (approved):
🚀

PR #258, opened 2022-07-19
Cast components to an array to prevent method calls on nil

njseeto:
We have recently received a handful of [Sentry errors](https://sentry.io/organizations/ministryofjustice/issues/3397888992/events/?project=5684538&referrer=slack) which have been due to methods being called on nil, meaning no components exist on a page.
Casting to an Array means we now call methods on an empty array if no components exist on a page.


brenetic (approved):
🚀

PR #259, opened 2022-07-19
Add AutocompleteItem model and ability to assign items to autocomplete components

njseeto:
[Trello](https://trello.com/c/82s692KP/2714-autocomplete-preview-mechanism)

### Add AutocompleteItem model
This adds the model which maps the metadata structure of a given autocomplete item to the corresponding id and name methods that are used when generating the select component on the page.

### Instantiate AutocompleteItem objects when autocomplete component
When the component is an autocomplete type then we want to make use of the new AutocompleteItem model instead of the the standard Item model.

### Assign autocomplete items for autocomplete components in preview
Since the items are not readily available in the service metadata we need to call the Metadata API to find the correct items and assign these items to the correct component when we load an autocomplete page in preview mode.

### Add autocomplete as a supported component for a single question page
This enables autocomplete component and associated answer to render on the check your answers page

### In Preview mode
Screenshot 2022-07-19 at 16 49 09

Screenshot 2022-07-19 at 16 49 24

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,13 @@ All notable changes to this project will be documented in this file.
 ### Changed
 ### Fixed
 
-## [Unreleased]
+## [2.17.10] - 2022-07-20
+
+### Added
+
+- Add `AutocompleteItem` model, this maps the metadata structure of an autocomplete item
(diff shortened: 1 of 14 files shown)


chrispymm (approved):
All looks good to me

PR #260, opened 2022-07-22
Return both text and value when calling id on an AutocompleteItem

njseeto:
## Update autocomplete page and select schemas

The autocomplete page needs to have the items property.

The select schema does not need to have the items property.

The select_item schema needs to have both text and value properties.

## Return both text and value when calling id on an AutocompleteItem

When the items for a component are passed into an autocomplete page
template `id` is used to set the 'value' on each select item. When a
user chooses and item and submits the answer it this value that gets
sent to the datastore.

However when we come to present the answer back to the user, either on
the autocomplete page when they go back to it or on a checkanswers page,
we want to have the text (aka name) as this is usually the human
readable form.

When we come to submit the form payload at that point we want to have
the 'value' since this is the thing that a Form Owner would actually
make use of.

Also add a countries fixture for use with testing.

## Override to_json method on the component model

Calling to_json on a component objects injects all the component
attributes into the data-fb-content-data element for that component.
This is updated whenever a user makes changes to an interactable element
in the Editor, such as a component question or a section heading etc.

For autocomplete components we do not need to inject all the select
items into the data-fb-content-data element as they are not interactable
in the Editor. Therefore we can just remove them entirely before loading
the page.
 
## Publish 2.17.12

### DIFF
--- a/app/models/metadata_presenter/component.rb
+++ b/app/models/metadata_presenter/component.rb
@@ -25,13 +25,17 @@ class MetadataPresenter::Component < MetadataPresenter::Metadata
   end
 
   def item_klass
-    type == 'autocomplete' ? MetadataPresenter::AutocompleteItem : MetadataPresenter::Item
+    autocomplete? ? MetadataPresenter::AutocompleteItem : MetadataPresenter::Item
   end
 
   SUPPORTS_BRANCHING = %w[radios checkboxes].freeze
 
   def supports_branching?
     type.in?(SUPPORTS_BRANCHING)
+  end
+
+  def autocomplete?
+    type == 'autocomplete'
   end
 
   def content?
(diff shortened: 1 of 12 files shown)


chrispymm (approved):

app/models/metadata_presenter/component.rb
```diff
@@ -10,6 +10,17 @@ class MetadataPresenter::Component < MetadataPresenter::Metadata
...
 
+  # Overriding here because autocomplete component's items property is non interactable
+  # in the Editor therefore it does not need to exist in the data-fb-content-data
+  # attribute on the page. The data-fb-content-data attribute is what is used to
+  # update user interactable elements such as the component question or page section
+  # heading etc.
+  def to_json(*_args)
+    return super unless autocomplete?
+
+    JSON.parse(super).to_json(except: 'items')
```
chrispymm, line 21: This is definitely cleaner!  Although I find it strange that you can't just do `super(except: 'items')` 🤷‍♂️

chrispymm (approved):
🚀

chrispymm (approved):
🚀

PR #262, opened 2022-07-28
Add autocomplete component

chrispymm:
PR to add the govuk autocomplete component into the presenter.

Includes adjustments to loading of styles scripts. 

Moved JS to head with `defer` attribute for better performance (being in the head means it gets discovered early by the browsers parser / preload scanner, but defer means its till runs after HTML is in place)


brenetic (approved):
🚀

PR #263, opened 2022-08-03
Validates when user's answer is not in autocomplete list

njseeto:
[Trello](https://trello.com/c/5sfGGfLx/2718-autocomplete-runner-validations)
 
### Add autocomplete items validator
This validator compares the user's answers to the autocomplete item select list. If the user's answer is not on the list, the answer is invalid.

### Pass `autocomplete_items` through to the validator
We need the autocomplete items in the validators so we can compare the user's answers to the items list on the component. We experimented with adding a module in the Runner and Editor apps that would allow us access to the items, however, this left us with the inability to sufficiently test the implementation. So we decided to go with passing the items in from the answers controller.
 
### Ensure the autocomplete items list is available in the answers and pages controllers
We require this method in both controllers as a user could find themselves on the autocomplete component on initial page load (GET) or after their answer has been validated (POST). Without this, the autocomplete select list would not be loaded into the component.

### 2.17.14

### DIFF
--- a/app/validators/metadata_presenter/autocomplete_validator.rb
+++ b/app/validators/metadata_presenter/autocomplete_validator.rb
@@ -1,0 +1,21 @@
+module MetadataPresenter
+  class AutocompleteValidator < BaseValidator
+    attr_reader :autocomplete_items
+
+    def initialize(page_answers:, component:, autocomplete_items:)
+      super(page_answers: page_answers, component: component)
+
+      @autocomplete_items = autocomplete_items
+    end
+
+    def invalid_answer?
+      return if autocomplete_item_list.blank?
+
+      autocomplete_item_list.exclude?(JSON.parse(user_answer))
+    end
+
+    def autocomplete_item_list
+      @autocomplete_item_list ||= autocomplete_items[component.uuid]
+    end
+  end
+end
--- a/default_metadata/component/autocomplete.json
+++ b/default_metadata/component/autocomplete.json
@@ -7,6 +7,7 @@
   "name": "component-name",
   "legend": "Question",
   "validation": {
-    "required": true
+    "required": true,
+    "autocomplete": true
   }
 }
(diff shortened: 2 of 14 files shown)


app/validators/metadata_presenter/autocomplete_item_validator.rb
```diff
@@ -0,0 +1,19 @@
...
+    attr_reader :autocomplete_items
+
+    def initialize(page_answers:, component:, autocomplete_items:)
+      super(page_answers: page_answers, component: component)
+
+      @autocomplete_items = autocomplete_items
+    end
+
+    def invalid_answer?
+      autocomplete_item_list.blank? ? true : autocomplete_item_list.exclude?(JSON.parse(user_answer))
```
brenetic, line 12: we want to avoid returning primitive boolean i think, especially as our other functional calls will already perform this for us. essentially you can just `return` with a guard clause for any times that you need which will mean falsey.

```
return if autocomplete_items[component.uuid].blank?

autocomplete_item_list.exclude?(JSON.parse(user_answer))
```

no primitive return needed then

default_metadata/component/autocomplete.json
```diff
@@ -7,6 +7,7 @@
   "name": "component-name",
   "legend": "Question",
   "validation": {
-    "required": true
+    "required": true,
+    "autocomplete_item": true
```
brenetic, line 11: i think we should just use `autocomplete` that is the name of the component. also there is never just one item

app/validators/metadata_presenter/autocomplete_validator.rb
```diff
@@ -0,0 +1,21 @@
...
+    end
+
+    def invalid_answer?
+      return if autocomplete_item_list.blank?
+
+      autocomplete_item_list.exclude?(JSON.parse(user_answer))
+    end
+
+    def autocomplete_item_list
+      autocomplete_items[component.uuid] || []
```
brenetic, line 18: you can memoize this. also since you now have the guard clause there is no need to set the `[]` as an alternative as the guard clause will return if it is `nil`.

brenetic (approved):
🚀

PR #265, opened 2022-08-08
Add ability to manually inject properties into Global Analytics

stevenburnell-moj:
Updated Global Analytics to allow for develop injection of key:value pairs to the GA4 initiation process. 
This was originally designed to allow for manual page title values, like this:

gtag('config', 'MEASUREMENT_ID', {
   'page_title': 'Travel Destinations',
   'currency': 'USD'
});

(example taken from Google Documentation)

The implementation allows for more than the page_title addition alone, since you could also inject the currency value, if desired. Values added can come from the related page type template (in this commit that means the Confirmation page) but also allows for putting values directly into the global analytics template. 

Setting as a JS variable rather simply adding direct to the gtag function should mean developers can test out value using the browser console, without having to keep updating the Metadata Presenter templates, until ready.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ All notable changes to this project will be documented in this file.
 ### Changed
 ### Fixed
 
+## [2.17.16] - 2022-08-08
+
+### Added
+
+ - Ability to manually inject properties into Global Analytics
+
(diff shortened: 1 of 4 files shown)


brenetic (approved):
🚀

PR #269, opened 2022-08-25
Presenter specific locales

brenetic:
In order to prevent accidental clashes of property names when the
presenter is used in either the Runner of the Editor move the locales
under a `presenter` property.


H3ll3m4 (approved):
🚀

H3ll3m4 (approved):
🚀

PR #271, opened 2022-09-09
2.17.21 Update static Cookie content to be included in content lookups

stevenburnell-moj:
Correction to PR https://github.com/ministryofjustice/fb-metadata-presenter/pull/270 which didn't added correct content type. Not changing version because the intention of this change is the same as previous (last) commit and haven't updated Editor or Runner to pull in that (2.17.21) version, yet.

### DIFF
--- a/app/views/metadata_presenter/page/standalone.html.erb
+++ b/app/views/metadata_presenter/page/standalone.html.erb
@@ -9,7 +9,7 @@
          <%= t('presenter.footer.cookies.heading') %>
        </h1>
 
-       <div data-fb-content-type="content">
+       <div data-fb-content-type="static">
          <%= to_html t('presenter.footer.cookies.body') %>
        </div>
 


brenetic:
will need to bump the version in order for the changes to be picked up by the runner and editor

stevenburnell-moj:
As per description, this is a correction to the previous (merged) PR which changed version to 2.17.21. 

Because Editor and Runner have not yet been updated from 2.17.20, they are currently blissfully unaware of the screw-up. Have not changed the version due to this and the fact 2.17.21 (with this corrected change) will still be as per the change log.

brenetic:
the change in the version number is what triggers the publishing of the gem to RubyGems. The published version on RubyGems is what the Editor and Runner use

stevenburnell-moj:
Closing to replace with cleaner PR https://github.com/ministryofjustice/fb-metadata-presenter/pull/272

PR #275, opened 2022-11-22
Adding reference number to the confirmation page

njseeto:
[Trello](https://trello.com/c/pXXDM00K/3075-reference-number-adjust-template-in-the-presenter)

### Add reference number to confirmation page template
This commit adds the reference number to the confirmation page template.
The reference number will be non-editable and needs to be auto-generated on submission.
We tested the 'REFERENCE_NUMBER' env vars locally, so can confirm the reference number will only appear when the env var is enabled.

### Show the correct reference number dependent on the app
When in the Editor app, there is no need to generate a reference number, so we want to use a placeholder.

When in the Runner, we want to generate a new reference number every time.

`reference_number_enabled?` and `show_reference_number` are methods that will be defined in the Editor and the Runner.

### Bump to version 2.17.24

Screenshot 2022-11-22 at 11 23 54

### DIFF
--- a/app/views/metadata_presenter/page/confirmation.html.erb
+++ b/app/views/metadata_presenter/page/confirmation.html.erb
@@ -7,24 +7,23 @@
       <%= @page.heading %>
     </h1>
 
-  <% if ENV['REFERENCE_NUMBER'] == 'enabled' %>
-    <div class="govuk-panel__body">Your reference number</div>
     <div class="govuk-panel__body">
-      <strong>
-      <%= @page.reference_number %>
-      </strong>
+      <% if reference_number_enabled? %>
+        <p>Your reference number<br />
+          <strong><%= show_reference_number %></strong>
+        </p>
+      <% end %>
+
+      <%# Do not replace with app/views/metadata_presenter/attribute _lede.html.erb because that is different %>
+      <% if @page.lede %>
+        <div class="fb-editable"
+            data-fb-content-id="page[lede]"
+            data-fb-content-type="element"
+            data-fb-default-text="<%= default_text('lede') %>">
+          <%= @page.lede %>
+        </div>
+      <%- end %>
     </div>
-    <br>
-  <% end %>
-    <%# Do not replace with app/views/metadata_presenter/attribute _lede.html.erb because that is different %>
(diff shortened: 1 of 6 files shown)


app/views/metadata_presenter/page/confirmation.html.erb
```diff
@@ -7,6 +7,16 @@
...
+    <% if reference_number_enabled? %>
+      <div class="govuk-panel__body">Your reference number</div>
+      <div class="govuk-panel__body">
+        <strong>
+          <%= show_reference_number %>
+        </strong>
+      </div>
+      <br>
+    <% end %>
+
```
chrispymm, line 19: We really only want one `govuk-panel__body`class.  So probably a bit more like this:

```
<div class="govuk-panel__body">
  <% if reference_number_enabled? %>
    <div>Your reference number</div>
    <div>
      <strong>
        <%= show_reference_number %>
      </strong>
    </div>
    <br>
  <% end %>

  <%# Do not replace with app/views/metadata_presenter/attribute _lede.html.erb because that is different %>
  <% if @page.lede %>
    <div class="fb-editable"
           data-fb-content-id="page[lede]"
           data-fb-content-type="element"
           data-fb-default-text="<%= default_text('lede') %>">
      <%= @page.lede %>
    </div>
  <%- end %>
</div>
```

I presume the `<br>` is for spacing?  You could try and replace the reference number related `<div>`s with `<p>` tags, which would be more semantic, but I'm not certain how that will render. Happy to pair to refine it a bit if needed...

chrispymm (changes requested):

app/views/metadata_presenter/page/confirmation.html.erb
njseeto: All done!

njseeto: Yep, the `<br>` is for spacing - will give the `<p>` a try

njseeto: Thanks Chris - all done

chrispymm (dismissed):
🚀

chrispymm (approved):
🚀

PR #276, opened 2022-12-01
Fix MaxSizeValidator human_max_size error on string values

chrispymm:
Fixes an error caused when the upload component max size validator is provided with a string value (which is in the default metadata for the component)

Adds test to ensure fix.


brenetic (dismissed):
🚀

brenetic (approved):

PR #277, opened 2022-12-05
Update H3 to H2 for Check your answers template

chrispymm:
Updates the Check your answers page template to use H2 tag for headings of muliple question pages.  

This gives no visual change, as the same `govuk-heading-m` class is used, but it makes the heading heirarchy of the page correct and avoids the 'skipped heading level' warning from accessibility audits.


brenetic (approved):
🚀

njseeto (approved):
🚀

PR #280, opened 2022-12-16
Adds continue to pay button on confirmation page if payment_link is enabled

chrispymm:
If payment link is enabled we need to show a 'continue to pay' button on the confirmation page to enable the user to take action.

Runner:
image

Editor (Edit screen):
image

Editor (Preview):
image


brenetic (approved):
🚀 

can we stick a screen shot into the PR message please

PR #281, opened 2022-12-19
Use relative or fully qualified links in footer

brenetic:
There may be a need in the future to allow for some of the footer URLs
to link to external domains. By not prepending the request.script_name
to the footer href we allow for the metadata link to be a fully
qualified link OR a relative link

If the relative link does not have a '/' at the start then prepend one.

Publish 2.17.30


njseeto (approved):
🚀

PR #282, opened 2022-12-20
Fix preview link path generation

brenetic:
The Editor is a multi tenant application. It allows the previewing of
several different forms. In those instances we need to include the relative
of the service that is being edited.

Publish 2.17.31


njseeto (approved):
🚀

PR #288, opened 2023-02-06
Bump ruby3

H3ll3m4:
Upgrade Ruby version to 3.1.3, remove ruby version from brakeman ignore list and publish presenter to version 2.17.35


app/validators/metadata_presenter/validate_answers.rb
```diff
@@ -22,9 +22,9 @@ def validators
       components.map { |component|
         component_validations(component).map do |key|
           "MetadataPresenter::#{key.classify}Validator".constantize.new(
-            {
-              component: component,
-              page_answers: page_answers
+            **{
```
StevenLeighton21, line 25: What does the ** do?

H3ll3m4: It specifies passing keyword arguments instead of a hash see https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/

StevenLeighton21 (approved):

PR #290, opened 2023-02-27
Fix formatting of content component on the checkanswers page

chrispymm:
The `@page.send_body` content component on the check answers page is an editable content component, however it is not rendered into the components array like all the other components.

There was a bug in how links were being rendered in this component. They were being handled incorrectly and link syntax was being escaped.

This was caused because we were outputtung markdown into the content of the comopnent on page render rather than convering to html first (like we do in all other components).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ All notable changes to this project will be documented in this file.
 ### Changed
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


StevenLeighton21 (approved):
🕵️

H3ll3m4 (approved):
🚀

PR #292, opened 2023-03-14
Update content to reflect 20 minute session

StevenLeighton21:
We previously had a 60 minute session timer to complete filling in the form, this has now changed to a 20 minute timer that resets on every API interaction, updating the cookie policy and expiry page to reflect this.


njseeto (approved):
🚀

PR #293, opened 2023-03-23
Initial pass - view and empty controller in place, other pages link t…

StevenLeighton21:
…o view

### DIFF
--- a/app/views/metadata_presenter/save_and_return/show.html.erb
+++ b/app/views/metadata_presenter/save_and_return/show.html.erb
@@ -1,0 +1,45 @@
+<div class="fb-main-grid-wrapper">
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+    <h1 id="page-heading" class="govuk-heading-xl"><%= t('presenter.save_and_return.show.heading') %></h1>
+    <p class="mojf-settings-screen__description"><%= t('presenter.save_and_return.show.description') %></p>
+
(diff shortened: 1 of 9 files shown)


app/views/metadata_presenter/save_and_return/show.html.erb
```diff
@@ -0,0 +1,43 @@
...
+          bold_labels: false
+        %>
+        <%=
+        f.govuk_text_field :secret_answer,
+          label: { text: t('presenter.save_and_return.show.secret_answer') },
+          name: "secret_answer"
+        %>
+      </div>
+
+      <%= f.govuk_submit t('presenter.save_and_return.show.continue') %><% end %> <%= button_to t('presenter.save_and_return.show.cancel'), :back, { class: 'govuk-button govuk-button--secondary'} %>
```
chrispymm, line 40: I think `button_to` may be problematic here.  Is there a specific reason you've used it?
`button_to` generates a form with a button in, which will result in a `form` within a `form` which is invalid html. 

This should, I think, semantically be a link as it navigating to a new location.  I appreciate that Fabien has styled it as a button, and I sort-of understand why, but you shouldn't really style links to look like buttons, and the design system by default doesn't allow it (we have overrides in the editor to do this, but not in the runner)

If you look at this example in the design system, https://design-system.service.gov.uk/components/button/#grouping-buttons, we are technically the second example, where the 'cancel' action is a link.  A secondary button should only be used if it actually a second action *on the form itself*

PR #294, opened 2023-03-24
Add session timeout warning modal

chrispymm:
This updates the views to add a session timeout warning modal.

image

The feature consists of a modal component wrapped within a session timeout component.

The modal component requires the rest of the page content to be wrapped within a containing div so that it can be made `inert` when the modal is shown.

The timeout component also has a non-js fallback element that is shown either if JS is disabled/not-loaded or if the brwoser does not support the `dialog` element or its  polyfill.

image


CHANGELOG.md
```diff
@@ -7,10 +7,17 @@ All notable changes to this project will be documented in this file.
 ### Changed
 ### Fixed
 
+## [2.17.40] - 2023-03-20
+### Added
+
+ - Added session timeout warning modal.  After 25 minutes of inactivity a
```
StevenLeighton21, line 13: The text we show users says 30 minutes, this says 25 :D

StevenLeighton21 (approved):

CHANGELOG.md
chrispymm: 30 minutes is the total session time.  25 mins of inactivity + 5 mins of warning modal = 30 mins

PR #295, opened 2023-03-24
Separate cookie request and cookie confirmation message

chrispymm:
### Problem:
When accepting cookies, the tracking code did not load until the next page load, so the tracking for the initial page is 'lost'

### Solution:
Reload the page on accept (and reject, just for consistency of impolementation).
When relaoding the page we add a querystring to indicate cookies accepted/rejected.

Previously the cookie request, and the accep, reject messages were all in the same template and just shown/hidden by JS.

Now we're lopading/showing them separately they need to be in different partials, and be added into the page under slightly different conditions.

**N.B.**
In order to avoide too many version bumps, I've snuck in one very small change to the session timeout warning:  I've updated the minutes idle to 24 instead of 25.  

The reason for this is that due to the way javascript works `setTimeout` and `setInterval` are not perfect in terms of timing.  This means that with a backend session timeout of 30mins and a frontend idletime of 25mins plus warning time of 5mins, if those timers are fractionally out, there is an edge case that if someone waited until the last couple of seconds to dismiss the modal and continue, their session may have already timed out on the backend 😬.  

An idle time of 24mins allows nearly 2s of leeway for each minute of JS timing we do, which should be fine (and no one is going to time the full idle time and countdown to discover the 1min discrepancy!)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ All notable changes to this project will be documented in this file.
 ### Changed
 ### Fixed
 
+## [2.17.43] - 2023-03-29
+### Changed
+
+ - Changed the templates for the cookie warning banner.  We reload the
+     page between the request and the accept or reject banner, so they need to
+     be separated into separate partials with different conditions for being
+     shown.
+
 ## [2.17.42] - 2023-03-29
 ### Changed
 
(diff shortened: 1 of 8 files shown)


njseeto (approved):
🚀  LGTM!

PR #296, opened 2023-03-28
Save and return - submit saved progress

StevenLeighton21:
Add controller and views for save and return up to submitting the confirmed email

TODO:
Send email to submitter
Clear session after submission
Error handling on POST to user-datastore
Fix back to form button to return to <form_url><page_slug> rather than :back

## Flow

Added show method that renders
image

Filling in the form creates a SavedForm model and validates it. Successfully passing validation and submitting the form saves the input to session and renders

image

The input email is validated against the previously entered email. On success we show a confirmation page 

image

The metadata presenter will pass the saved form progress to the runner engine, which will post to the fb-user-datastore for persistence.


config/locales/en.yml
```diff
@@ -48,6 +48,19 @@ en:
         secret_answer: 'The answer to your security question'
         continue: 'Continue with saving'
         cancel: 'Cancel saving and resume form'
+      confirm_email:
+        heading: 'Check your email address is correct'
+        description: 'Make sure this is correct before finishing'
+      saved:
+        heading: 'Your form has been saved'
+        success_1: 'We have sent a one-off link to '
+        success_2: ' which you can use to resume this form within 28 days. After that time, your saved information will be deleted.'
```
njseeto, line 57:
```suggestion
        success: "We have sent a one-off link to %{email} which you can use to resume this form within 28 days. After that time, your saved information will be deleted."
```
Could we interpolate the email into the locale? Then I _think_ this might work in the template:
`t('presenter.save_and_return.saved.success', email: confirmed_email)`

config/locales/en.yml
```diff
@@ -48,6 +48,19 @@ en:
...
+        info_1: "If you haven't received the email after a few minutes, check your spam folder."
+        info_2: 'This link will only work once. If you want to save your progress again after resuming the form, you will need to repeat the save process and generate another link'
+        info_3: 'You are now free to close this window or navigate away from this page.'
```
njseeto, line 60:
```suggestion
        info: 'If you haven't received the email after a few minutes, check your spam folder. <br />This link will only work once. If you want to save your progress again after resuming the form, you will need to repeat the save process and generate another link<br />You are now free to close this window or navigate away from this page.'
```
We could put the breaks in the locale, then we can have one `p` tag on the template:
`<p class="govuk-hint"><%= t('presenter.save_and_return.saved.info') %></p>`

spec/validators/email_confirmation_validator_spec.rb
```diff
@@ -0,0 +1,15 @@
+RSpec.describe SavedProgressValidator do
```
njseeto, line 1:
```suggestion
RSpec.describe EmailConfirmationValidator do
```

njseeto (changes requested):
looks good - couple of small comments

njseeto (approved):

njseeto (approved):

PR #298, opened 2023-03-29
Update in_progress? helper to allowed_page?

chrispymm:
The in_progress? method was incorrectly named - it was returning true for standalone pages, the start page, and the session expired url.

Have renamed this method to `allowed_page?` which returns to the previous language and is actually accurately describing the condition!

So now if we are not on an *allowed_page* we want to include the session timeout warning partials.


njseeto (dismissed):
🚀

PR #301, opened 2023-04-06
Add some error handling to POSTing saved progress and a spec

StevenLeighton21:


app/controllers/metadata_presenter/save_and_return_controller.rb
```diff
@@ -41,11 +41,12 @@ def confirm_email
...
 
       if @email_confirmation.valid?
-        # save_form_progress returns the uuid of the saved form as 'id'
-        # user to generate email link, for now not assigned because rubocop
-        # save_form_progress.body['id']
-        # TODO: rescue failed POST
-        # send_email(uuid, confirmation_params[:email_confirmation])
+        response = save_form_progress
+        if response.status != 200
+          render :email_confirmation, status: :unprocessable_entity and return
```
njseeto, line 46: A question about these `render: :unprocessable_entity` on line 46 and 52. The first one is when there is no match in the database (ie: the email did not match the session) and the one on line 52 is when the POST request fails altogether?

StevenLeighton21: Yes, it should probably be something other than unprocessable entity, though I'm not really sure. We're in 'should never happen' territory, maybe the best thing to do is redirect to the 'something went wrong' page, or redirect back to the page they left off from with a message about it failing to save progress.

njseeto (approved):

PR #304, opened 2023-04-12
WIP

StevenLeighton21:


app/views/metadata_presenter/save_and_return/save_progress.html.erb
```diff
@@ -1,13 +1,14 @@
 <div class="fb-main-grid-wrapper">
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
+    <div class="hidden" data-test=<%= show_reference_number %>/>
```
njseeto, line 4: I'm not sure we can do this because reference number is not required for save and return so i think this might be `nil`/blow up in the runner if reference number feature is disabled but save and return is enabled?

StevenLeighton21: Yeah you're right - it's completely unused so I have removed it. Was an experiment in accessing the ref for getting the email in the acceptance test :)

chrispymm (approved):
This all looks good to me, can't see any issues or things I'd change.

PR #307, opened 2023-04-21
Submitter save return

njseeto:
### Add ability to create a save and return submission
We will use a similar mechanism to the `create_submission`.

The `create_save_and_return_submission` will take a payload:

`{id: '1234567', email: 'email@email.com'}`

Where `id` is the id of the `SavedForm` record in the UserDatastore.

This will call the Runner `create_save_and_return_submission` method and utilise the existing submission mechanism.


StevenLeighton21:
Branch has been pulled into https://github.com/ministryofjustice/fb-metadata-presenter/pull/304 if you want to close this PR @njseeto :)

PR #310, opened 2023-04-26
Fix question heading spacing issues

chrispymm:
There were spoacing issues between the page title and the first component on multiple question pages when the first component was a checkbox or radion component.

This was caused due to the margin between components being added into the `label` element within the question,a nd the checkboxes and radios components have a `fieldset` element that was preventing the margin between the title and the question collapsing.

Solution was to move the margin out onto the actual question element itself in order for the margins to collapse correctly.

image
Before: Too much spacing betweeen title and first question

image
After: Margins collapsing correctly

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@ All notable changes to this project will be documented in this file.
 ### Fixed
 
 ## [2.18.4] - 2023-04-14
+### Fixed
+
+ - Fixed spacing between titles and components on multiple question pages.
+
+## [2.18.4] - 2023-04-14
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ All notable changes to this project will be documented in this file.
 ### Changed
 ### Fixed
 
+## [2.18.4] - 2023-04-14
```
njseeto, line 10:
```suggestion
## [2.18.5] - 2023-04-26
```

njseeto (approved):
🚀

PR #317, opened 2023-05-19
Suffix duplicate file upload

StevenLeighton21:
Adds a suffix to filename on save if it matches filename of a previous answer to ensure it is saved and included in attachments

image

### DIFF
--- a/app/controllers/metadata_presenter/answers_controller.rb
+++ b/app/controllers/metadata_presenter/answers_controller.rb
@@ -72,12 +72,12 @@ module MetadataPresenter
       @page_answers.page.upload_components.each do |component|
         answer = user_data[component.id]
 
-        if answer.presence
-          filename = answer.nil? ? "" : answer["original_filename"].delete('>"[]{}*?:|]/<')
-          extname = File.extname(filename)
-          basename = File.basename(filename, extname)
-          filename_regex = Regexp.new("^#{Regexp.quote(basename)}(?>-\((\d)\))?.#{Regexp.quote(extname)}")
+        original_filename = answer.nil? ? @page_answers.send(component.id)["original_filename"] : answer["original_filename"]
 
+        if original_filename.presence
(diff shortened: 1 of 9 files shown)


app/controllers/metadata_presenter/answers_controller.rb
```diff
@@ -72,6 +80,12 @@ def upload_files
       @page_answers.page.upload_components.each do |component|
         answer = user_data[component.id]
 
+        original_filename = answer.nil? ? @page_answers.send(component.id)['original_filename'] : answer['original_filename']
+
+        if original_filename.presence
```
njseeto, line 85:
```suggestion
        if original_filename.present?
```

Should we return a boolean rather than the `original_filename` object in this statement? Tbh I'm not sure if it amounts to the same thing

app/models/metadata_presenter/page_answers.rb
```diff
@@ -75,7 +74,16 @@ def raw_date_answer(component_id)
     def filename(path)
       return sanitize(path) if path.nil?
 
-      sanitize(path).gsub(/&gt;/, '').gsub(/&lt;/, '').delete('>"[]{}*?:|]/<').delete('\\')
+      filename = sanitize(path).gsub(/&gt;/, '').gsub(/&lt;/, '').delete('>"[]{}*?:|]/<').delete('\\')
+
+      if count.presence && count.positive?
```
njseeto, line 79:
```suggestion
      if count.present? && count.positive?
```

Similar to above comment.

njseeto (commented):
Great stuff! Just one small comment on `presence` vs `present?`

app/controllers/metadata_presenter/answers_controller.rb
StevenLeighton21: pretty much does, O think present is probably better?

njseeto (approved):
Looks good! 🚀

PR #320, opened 2023-05-22
Update confirmation page template

chrispymm:
Moves the additional content areas on the confirmation page inside the width container.

Before:
image

After:
image


H3ll3m4 (approved):
🚀

PR #323, opened 2023-05-24
Add submission complete page

chrispymm:
Adds a template for the submission complete page, which will be shown to the user if they try and use the back button after the confirmtaion page.

We also add a route for it.


H3ll3m4 (approved):
🚀

PR #326, opened 2023-06-20
Rename filename when extension is 'jfif' or 'jpg'

njseeto:
### Rename filename when extension is 'jfif' or 'jpg'

We have a user who relies on consolidating the uploaded file name of email attachments with the filenames in the users answers.

When a user uploads a file that has the extension of either 'jfif' or 'jpg', the mime type used is jpeg. This is because all these extensions are of mime type jpeg.
Our submitter uses the mime type as the extension to ensure the email attachment is not corrupt and can be downloaded.

This commit silently changes the filenames to ensure we use the same extension that will be created when a jfif or jpg image is uploaded.

### Bump to version 3.0.4


spec/models/page_answers_spec.rb
```diff
@@ -151,26 +151,47 @@
...
+            let(:expected_mimetype) { 'image/gif' }
+
+            it 'returns file details hash' do
+              expect(
+                page_answers.send('dog-picture_upload_1')
+              ).to include(expected_answer)
+            end
+          end
+
+          context 'when file extension should be renamed' do
```
StevenLeighton21, line 181: Incredibly picky comment so feel free to ignore but the bug we're fixing is JFIF and JPG to use JPEG, so might be nice to add test cases to cover both JFIF and JPG

StevenLeighton21 (approved):

StevenLeighton21 (approved):

PR #333, opened 2023-06-28
Determine if standaloine pages contain placeholder values

chrispymm:
Standalone pages are created with default templates that contain various bits of placeholder text that a form editor needs to relace with their own content e.g.

> [list the type of personal data used e.g. name, address, contact details etc]
>
>We also collect special categories of information, such as [delete this section if not >applicable]:

We need to be able to identify when a form editor has replaced all of these placeholder elements within the page body text.

This PR creates a lisat of these placeholder values within the translations file, and this is then used within a regex to scan the body text for matches.

### Is the translationms the right place for this list?  
It feels good for it to be adjacent to the actual content, but there is duplication which means when updating the body content, you have to be sure to update all the placeholders

### Could/should we use the placeholders as replacements?  
Theoretically, yes, we could do something like this:
```
#en.yml
body: "We also collect special categories of information, such as %{0} and %{1}"
placeholders:
  - "[delete this section if not applicable]"
  - "[more placeholder content]"

#html.erb file
t('presenter.footer.privacy.body', **(t('presenter.footer.privacy.placeholders').map.with_index {|v,i| [i,v]}.to_h})) 
```
However this is hard to read in code, and harder to read and write the translations (IMO)

### Could we eliminate the need for the 'placeholders' key entirely and be more dynamic?
Yes, but we'd need to change how we surround our replacements.  

Currently we use `[]` to mark our placeholders, however this is also valid markdown syntax for links, so we can't use a regex to simply check for the existence of [].  (Also it is within reason that someone could also legitimately use sinlge quare brackets in their content.

If we were to update our placeholders to something like `{{ }}` then we could use one regex to find if there were any within the content, without needing to explicitly check for their contents.

This would proibably require updating other areas where we use these type of replacement strings for consistency (e.g. in the email copy) and that might be non-trivial.


spec/models/page_spec.rb
```diff
@@ -359,4 +359,56 @@
       expect(expected_items.first.value).to eq('123')
     end
   end
+
+  describe '#contains_placeholders?' do
+    context 'when page has placeholders' do
+      subject(:page) { service.find_page_by_url('privacy') }
+
+      before do
+        allow(page).to receive(:placeholders).and_return(['[insert your organisation name]'])
```
StevenLeighton21, line 368: _super_ picky comment but it's probably worth having another item in this placeholder array to test your join in the regex generation

StevenLeighton21 (approved):

PR #335, opened 2023-07-06
Fix 404 when resuming form with custom CYA page

StevenLeighton21:
Save and return would fail if the form had a custom check answers page as it was being looked up by URL, where it should be looked up by type.

Working in dev on the pentest form
image


njseeto (approved):
🚀

PR #339, opened 2023-07-10
Multi file upload

StevenLeighton21:
Single upload page

image

Multiple files

image

Required validation

image

File type validation

image

Duplicate file validation

image

Limit reached

image

Can duplicate files from other pages, filenames will be appended

image

File names are sanitised

image

image

Check your answers!

image


app/controllers/metadata_presenter/answers_controller.rb
```diff
@@ -67,6 +116,16 @@ def render_validation_error
       render template: page.template, status: :unprocessable_entity
     end
 
+    def incoming_answer
```
StevenLeighton21, line 119: The multi upload component is the first to treat incoming answers as additional to the existing ones for that component. It's meant treating them as a special case in a lot of places, and if we implement further components that behave this way we might consider refactoring this.

app/models/metadata_presenter/page_answers.rb
```diff
@@ -53,6 +55,64 @@ def upload_answer(component_id, _count)
...
+      end
+
+      if answers.blank?
+        return nil
+      end
+
+      if answers.is_a?(Hash) # rendering only existing answers
+        if answers[component_id].blank?
+          return
+        end
```
njseeto, line 85:
```suggestion
       return if answers[component_id].blank?
```

njseeto (commented):
great stuff, just a couple of comments around refactoring

njseeto (approved):

PR #342, opened 2023-07-25
Component schema changes

njseeto:
## Add conditionals and expressions in the content component schema
As part of the conditional content visibility work, the content components will need to accept the `conditionals` key and, within that,`expressions` key.
Adding these into the content component schema in preparation for when we start injecting conditional metadata into these components.
Since the presence of `conditionals` or `expressions` attributes are not required in the content component schemas this should not affect existing content components.

## Remove the `body` attribute from Content Pages and Confirmation Pages
When a user creates either a Content page or a confirmation page, a default body attribute with the value '[Optional content]' is created. From the UI, this looks like a content component, but it is actually the body attribute.
Since it is not a content component in the metadata, any content in this 'body' attribute cannot be made conditional.

To reduce any confusion on these pages, the 'body' attribute has been removed so any newly created content pages or confirmation pages will not have the default '[Optional content]'. This will force the form editor to create a content component.

In terms of existing content or confirmation pages, if the form editor has not modified the 'body' (ie: it is still set as '[Optional content]') it will not be rendered. If the body attribute has been modified by the form editor, the content will be shown in both the runner and the editor.

### Content Page - Before
Screenshot 2023-07-25 at 12 44 02

### Content Page - After
Screenshot 2023-07-25 at 12 42 50

### Confirmation Page - Before
Screenshot 2023-07-25 at 12 52 42

### Confirmation Page - After
Screenshot 2023-07-25 at 12 53 26

## Bump to presenter version 3.2.0

### DIFF
--- a/schemas/component/content.json
+++ b/schemas/component/content.json
@@ -42,35 +42,7 @@
             ]
           },
           "expressions": {
-            "$ref": "#/definitions/expressions"
-          }
-        }
-      }
-    },
-    "expressions": {
-      "type": "array",
-      "items": {
-        "type": "object",
-        "properties": {
-          "operator": {
-            "type": "string",
-            "enum": [
-              "is",
-              "is_not",
-              "is_answered",
-              "is_not_answered",
-              "contains",
-              "does_not_contain"
-            ],
-            "page": {
-              "type": "string"
-            },
-            "component": {
-              "type": "string"
-            },
-            "field": {
-              "type": "string"
-            }
+            "$ref": "definition.expressions"
           }
         }
       }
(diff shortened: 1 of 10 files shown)


schemas/component/content.json
```diff
@@ -17,6 +17,63 @@
       "type": "string",
       "content": true,
       "multiline": true
+    },
+    "conditionals": {
```
StevenLeighton21, line 21: Is the schema for conditionals re-usable from branching? If so do we have those we can reference, if not, part of me wants to call them something like content_visibility_conditions or something just to be explicit

StevenLeighton21 (dismissed):

StevenLeighton21 (approved):

PR #347, opened 2023-08-10
Save and return invalid date fix

StevenLeighton21:
Saving progress skips validation, so if an invalid date is saved it will cause a 500 error when rendering the resume progress page.
This prevents the error with an empty string.


njseeto (approved):
🚀

PR #348, opened 2023-08-11
Inform form filler of the confirmation email

H3ll3m4:
[Trello](https://trello.com/c/SzV8RNaW/3524-inform-form-filler-of-the-confirmation-email)

This is a really simple change to inform the form filler of the use of confirmation email.
We are adding a disclaimer for the email input page and the check your answer page. 

On the email input we warn the user the email will be used:
Screenshot 2023-08-11 at 16 20 38

For the check your answer page, we warn the user depending the confirmation email is enabled or not:
Screenshot 2023-08-11 at 16 22 12


app/views/metadata_presenter/component/_email.html.erb
```diff
@@ -9,3 +9,12 @@
     spellcheck: "false",
     autocomplete: "email"
 %>
+
+<% if is_confirmation_email_question? %>
+  <div class ="govuk-warning-text">
+    <span class="govuk-warning-text__icon" aria-hidden="true">!</span>
+    <strong class="govuk-warning-text__text">
```
StevenLeighton21, line 16: I think a space is needed here between the span and the text? I think we use the govuk margin class elsewhere,  add `govuk-!-margin-right-2` to the span class? It's not present in the example but in the screenshot the icon looks too close to the text :D

StevenLeighton21 (approved):
Looks good - approved with minor comment about the visual

app/views/metadata_presenter/component/_email.html.erb
H3ll3m4: Sorry this was solved at some point. This is how it looks like now: image

StevenLeighton21 (approved):

PR #356, opened 2023-11-02
Update json-schema gem

zheileman:
Use latest `json-schema` version (4.1.1).

Since version 3.0.0, the gem also performs validation of `const` attributes.

Attribute `_type` (const) in _radios_, _checkboxes_ and _date_ components was being wrongly validated against a
  duplicated `_type` inherited from `definition.fieldset`. Removed this attribute from the fieldset definition.

Added missing `page.exit` to allowed page types in base schema.

Fixed warning "character class has '-' without escape" in a pattern regexp.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,8 +2,7 @@
 log/*.log
 pkg/
 
-spec/dummy/db/*.sqlite3
-spec/dummy/db/*.sqlite3-journal
+spec/dummy/db/*.sqlite3*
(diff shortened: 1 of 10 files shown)


StevenLeighton21 (approved):

H3ll3m4 (approved):
🚀

StevenLeighton21 (approved):

PR #359, opened 2023-11-21
Fix several validations affecting date and upload components

zheileman:
This is a bunch of fixes for a few validation problems reported in sentry alerts:

**Fix date validator not raising errors for some invalid dates**
The validator was using `Date.strptime` which is kind of lenient on casting a date, even if the date was clearly wrong.

For example 315/12/2025 will be valid.

Changed to use `Date.civil` which is more strict, and also add a couple additional year validations to catch fat finger inputs.

**Implemented a catch-all validator**
This currently only have an effect for the `upload` and `multiupload` components which are the ones producing as far as I can tell these unhandled errors.

This should fix the sparse sentry errors: `no implicit conversion of Platform::FilestoreError into Hash`

Also, this can be used in the future with other components easily.

**Fixed missing filename in virus scan validator**
Before:
Screenshot 2023-11-21 at 09 30 47

After:
Screenshot 2023-11-21 at 09 39 47


app/validators/metadata_presenter/date_validator.rb
```diff
@@ -1,13 +1,18 @@
 module MetadataPresenter
   class DateValidator < BaseValidator
     DATE_STRING_VALIDATIONS = %w[date_after date_before].freeze
+    YEAR_LOWER_BOUND = 1000
+    YEAR_UPPER_BOUND = 2099
```
StevenLeighton21, line 5: can we do 3000? Just in case someone has a form for dealing with long land leases or something :D

config/locales/en.yml
```diff
@@ -115,7 +115,7 @@ en:
           continue: 'Continue'
         progress:
           heading: '%{service_name}'
-          content_1: 'You have sucessfuly retrieved your saved information.'
+          content_1: 'You have successfully retrieved your saved information.'
```
StevenLeighton21, line 118: 🙏

app/validators/metadata_presenter/date_validator.rb
zheileman: Oh yeah that makes sense!
Will change in a moment. I need to raise another PR in the filestore to go along with these changes 👍

default_metadata/string/error.catch_all.json
```diff
@@ -0,0 +1,6 @@
+{
+  "_id": "error.catch_all",
+  "_type": "string.error",
+  "description": "A generic validation error not handled by a more concrete validator",
+  "value": "We could not save your answer, please try again"
```
StevenLeighton21, line 5: I do feel like this message could offer something to users - we couldn't process that file, possibly because of its type - etc, this really needs some input from content/design though so I would ship it as is and bring this to attention for a future tweak

StevenLeighton21 (approved):

default_metadata/string/error.catch_all.json
zheileman: Yep, I get what you're saying. Thing is this is a catch-all and generic validator on purpose. If we want to have specific error messages for each of the different possible component errors then a catch-all is not the solution, we have to implement concrete validators, like we have now for virus scan, max size, etc. so we can customise each and all of their error messages.

It's true currently we are only using this for upload errors (the catch all validator checks the component is upload or multiupload) so we could reword the error to talk specifically about file uploads. But the idea was for this to be a fallback, catch-all generic validator with a generic error message instead of blowing up (like now) and giving the "something went wrong" page 😄

PR #367, opened 2024-01-10
Redirect to first question if using external start page

StevenLeighton21:
Support external start page configuration for forms

If the external start page is configured in editor, the URL for that page will be injected as an ENV var.

If this is present, the header link will point to the configured URL.
If this is present, the static start page will redirect to the first question
If in editor preview and using an external start page, there is an info box
If on the first page and showing the back link, and using an external start page, we link back out to that external URL

### DIFF
--- a/app/views/metadata_presenter/header/show.html.erb
+++ b/app/views/metadata_presenter/header/show.html.erb
@@ -26,17 +26,11 @@
         </span>
       </a>
     </div>
-    <% if use_external_start_page? && service.service_name.present? %>
+    <% if service.service_name.present? %>
       <div class="govuk-header__content">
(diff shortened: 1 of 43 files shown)


app/views/metadata_presenter/header/show.html.erb
```diff
@@ -26,7 +26,12 @@
         </span>
       </a>
     </div>
-    <% if service.service_name.present? %>
+    <% if use_external_start_page? && service.service_name.present? %>
```
zheileman, line 29: Would it make sense instead of this if-else, having a helper method, for instance `#start_page_url`, that returns the external url if `use_external_start_page?` otherwise returns `root_path`?
As the markup etc is the same and only the url changes, to avoid repetition. Just a suggestion!

zheileman (approved):

PR #369, opened 2024-01-17
Implement address postcode format validation

zheileman:
First pass on a postcode format validator to complement the address mandatory fields validator.

It uses a gem. There's a distinction between validity and existence. The gem validates the format of a postcode, but not whether it actually refers to a location.


StevenLeighton21 (approved):
Looks good to me!

PR #370, opened 2024-01-23
New gem version with address component

zheileman:
This PR bumps the version of the gem in preparation for release of the address component.

The new version will be able to be used in editor and runner safely, as address component will be put behind a feature flag on the editor for the time being until we are ready to release.

This PR also includes a few minor tweaks raised as part of the QA testing.


H3ll3m4 (approved):
🚀

PR #377, opened 2024-02-07
Protect from empty validation schema attribute

zheileman:
https://trello.com/c/q2iVyM9D

There is an edge case where in some forms the metadata JSON attribute `validation` is empty (empty hash).

The way the metadata is queried in the editor made this empty attribute provoke a lookup in the `MetadataPresenter::DefaultText` map which is intended for hint text, labels, etc. but not for "validation" (which is not present in the map)

This returned a nil, which made some things bad down the line in the editor.

Instead we are now returning the original value (in this case an empty hash) if there is no default text for it (as is the case for `validation` key).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,13 @@ All notable changes to this project will be documented in this file.
 ### Changed
 ### Fixed
 
-## [3.3.12] - 2024-01-29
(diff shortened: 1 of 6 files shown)


H3ll3m4 (approved):
🚀

PR #378, opened 2024-02-08
Sort components by order

chrispymm:
This PR adds the `Comparable` module to the `Component` class.  This allows us to define the `<=>` method to allow comparaison and sorting of component instances.

This makes ensuring the components are ordered correctly simple, as we can just call `sort` on an array of components.

If there is no order property for the component we return `0` meaning the order will remain unchanged.


StevenLeighton21 (approved):

zheileman (approved):
Awesome! 👍

PR #380, opened 2024-02-09
Basic auth replacement

zheileman:
https://trello.com/c/kgK7tVzy

This will "hook" into the current `require_basic_auth` runner before filter so it is as "drop-in" as possible.

It replaces the native browser basic auth with a hosted view / controller and form object to perform the authentication using the same user/password coming from ENV variables as before.

It sets a session cookie upon successful sign in and checks this cookie on each request to know if the user is allowed in.

A warning message is shown unless form is running in `live-production` or `test-production`.

Pending welsh translation of the new strings used in this feature. For the time being English strings will be shown.

Screenshot 2024-03-13 at 12 30 33

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ All notable changes to this project will be documented in this file.
 ### Changed
 ### Fixed
 
+## [3.3.25] - 2024-03-14
+### Added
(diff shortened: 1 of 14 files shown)


StevenLeighton21 (approved):

H3ll3m4 (approved):
🚀

PR #381, opened 2024-02-12
Rationalise save and return controller

zheileman:
Split the massive `SaveAndReturnController` into 2 separate controllers by leaving the functionality that logically belongs to the "save" process into the original controller, but shifting everything that belongs to the "return/resume" process to a separate `ResumeController`.

Additionally some common functionality has been shifted to a concern, although I'm sure a lot more could be shifted with a bit of refactor but didn't want to fundamentally change how this is done an the reasoning behind all this.

The ultimate objective of this refactor (apart from having code that is easier to understand and maintain) is to be able to introduce authorisation in one controller but not in the other controller more easily from the runner side.


app/views/metadata_presenter/resume/return.html.erb
```diff
@@ -1,10 +1,13 @@
 <div class="fb-main-grid-wrapper">
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
-    <h1 id="page-heading" class="govuk-heading-xl"><%= t('presenter.save_and_return.resume.heading', service_name: get_service_name) %></h1>
-    <p><%= t('presenter.save_and_return.resume.content') %></p>
+
     <%= form_for @resume_form do |f| %>
       <%= f.govuk_error_summary(t('presenter.errors.summary_heading')) %>
+
+      <h1 id="page-heading" class="govuk-heading-xl"><%= t('presenter.save_and_return.resume.heading', service_name: get_service_name) %></h1>
```
zheileman, line 8: Moved the h1 to conform to the design system guidelines for error summary:
`Put the error summary at the top of the main container. If your page includes breadcrumbs or a back link, place it below these, but above the <h1>.`

PR #384, opened 2024-02-26
Move components on a multiquestion page

chrispymm:
This PR updates the temp[lates to accomodate moving components on a multiquestion page

Changes include:
* Adding data attributes for stimulus controllers used in the editor
* Adding a wrapper div around editable content in order for it to work with move


zheileman (approved):
🎉

PR #388, opened 2024-03-04
Resolve issue with if/and conditions for content

StevenLeighton21:
This resolves a bug where if you have a condition that looks like `[if, and, and, and]` then it would hit the code branch for `[0] == if` and return `.any?` which meant that if was true, but none of the and conditions were, it would still be considered true and show the content.

New AT with a test for this in dev https://github.com/ministryofjustice/fb-acceptance-tests/pull/398
image


H3ll3m4 (approved):
🚀

PR #391, opened 2024-03-11
Regex validator

H3ll3m4:
### Context
This is part of the regex validation work. See ticket: 
[Trello](https://trello.com/c/mxZUs7H3/3949-add-regex-validator-to-the-presenter)

### Implementation
We want to add regex validation to string. There is already a the object pattern in the schemas used for strings so we are using it for text validation. 

We are adding a regex validator who just match the regex with user answers and check if it's valid.

### Tests
We are adding unit tests spec. 
Schema is valid see deployment on [Circle](https://app.circleci.com/pipelines/github/ministryofjustice/fb-editor/9953/workflows/6a3b65a6-84f0-4bff-9e47-9842f2543b87)

And it has been tested manually on the Editor see it working on single question page : 

Screenshot 2024-03-11 at 12 48 16

And multi question page too:

image


StevenLeighton21 (approved):
LGTM

PR #392, opened 2024-03-20
Support ampersand in autocomplete answers

StevenLeighton21:
This previously failed because the user answer would be converting & to &amp; - we can just gsub that before we compare to the autocomplete items without otherwise interfering with the answers.

To be deployed and tested to ensure autocomplete answers with & can be parsed into a submission etc

image

Fixed in dev runner
image (11)

Value in PDF output, email and csv
image
image
image


H3ll3m4 (commented):
Do we want to bump up the version too? 
Is there a ticket for that?

chrispymm (dismissed):
🚀

H3ll3m4 (approved):
🚀

PR #396, opened 2024-04-05
Back link nil guard

StevenLeighton21:
Going direct to a bad url (perhaps an old bookmark) would throw instead of returning a 404

Added a guard to prevent this

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ All notable changes to this project will be documented in this file.
 ### Changed
 ### Fixed
 
+## [3.3.31] - 2024-04-05
+### Fixed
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ All notable changes to this project will be documented in this file.
 ### Changed
 ### Fixed
 
+## [3.3.31] - 2024-04-05
+### Fixed
+ - Fixed issue where going directly to a 404 page could cause a vil value exception
```
zheileman, line 12: `vil` -> `nil` 😄

zheileman (approved):

zheileman (approved):

PR #398, opened 2024-04-17
Accessible labelling page editing

chrispymm:
This PR changes the presenter templates to enable better accessible labelling within the editor, and to address issues raised by the AXE audit tool.

* Fixes heading hierarchy issues
* Ensures editable headings use a child span as the editable element not the heading element itself.
* Ensures buttons are `aria-disabled` not `disabled` to enable them to be discoverable by AT users.


H3ll3m4 (approved):
🚀

PR #424, opened 2026-04-16
CDPTKAN-924: Fix snyk vunerabilities

ripan:
Update rails from 7.2.0 => 7.2.3.1
Update ruby from 3.2.0 => 3.4.9 ([stable release](https://www.ruby-lang.org/en/downloads/))

JIRA: https://dsdmoj.atlassian.net/browse/CDPTKAN-924


app/models/metadata_presenter/page_answers.rb
```diff
@@ -32,7 +32,7 @@ def method_missing(method_name, *_args)
         upload_answer(component.id, count)
       elsif component && component.type == 'multiupload'
         answer_object = multiupload_answer(component.id, count)
-        answer_object.to_h if answer_object.present?
+        answer_object.presence&.to_h
```
mseedat-moj: Is returning `nil` ok? the other if conditions all seem to return an empty hash from `upload_answer` I think?

mseedat-moj (changes requested):
Should always return empty hash? or is returning `nil` ok?

app/models/metadata_presenter/page_answers.rb
ripan: Good catch. I need to revert this rubocop suggestion. Rubocop is assuming answer_object as string hence applying presence on it.  Disabling rubocop suggestion

mseedat-moj (approved):

mattempty (approved):
🚀

PR #425, opened 2026-05-13
CDPTKAN-772: Fix AppSec issues

ripan:
- Cookie httponly is not set explicitly and defaults to false. Set it to true to protect the cookie from malicious client-side JavaScript code.
- Cookie secure is not set explicitly and defaults to false. Set it to true to ensure the cookie is only sent via HTTPS and help protect from man-in-the-middle attacks.

JIRA: https://dsdmoj.atlassian.net/browse/CDPTKAN-772
SNYK: https://app.snyk.io/org/central-digital/project/497d8b63-f6c1-4fec-b409-8d3fabb36661


mattempty (approved):
🚀

PR #426, opened 2026-05-14
Bump version to 3.4.14

ripan:
- Cookie httponly is not set explicitly and defaults to false. Set it to true to protect the cookie from malicious client-side JavaScript code.
- Cookie secure is not set explicitly and defaults to false. Set it to true to ensure the cookie is only sent via HTTPS and help protect from man-in-the-middle attacks.


wilson1000 (approved):
🚀

